PR #3, opened 2019-11-30
refactor: structure files and definitions symmetrically

ayazhafiz:
Moves `Token` to its own source file, renames `Data.hs` to `Lexer.hs` to
better represent the lexing functions in there, and gets rid of the
qualified `hgbLex` function in favor of `lex`.


src/Lexer.hs
```diff
@@ -1 +1,59 @@
...
+    -- Parse content after the opening string delimiter, returning as much as possible of
+    -- 1. String content
+    -- 2. String closing delimiter
+    -- 3. Tokens after string
+  where
+    (contentStr, postContentStr) = break (== delim) xs
+    content =
+      case contentStr of
+        [] -> []
+        otherwise -> [(Token Symbol.String contentStr)]
```
lukasberglund: I think it would be better if we simply created a String token with an empty string. I think that would make converting the tokens into a tree a lot easier.

src/Lexer.hs
```diff
@@ -1 +1,59 @@
...
+    postContent =
+      case postContentStr of
+        [] -> []
+        (endDelim:remainder) ->
+          (Token Symbol.StrBound [endDelim]) : lex remainder
```
lukasberglund: This looks good.

src/Lexer.hs
ayazhafiz: sounds good to me

PR #7, opened 2019-12-01
refactor Lexer and Token constructs

lukasberglund:
- Added beginning of grammar
- Changed name of variable in lexer.hs
- Changed order of lines in Symbol.hs
- Named the fields in the token data
- Changed test to reflect change in token


lukasberglund:
Okay done!

PR #9, opened 2019-12-03
Changed lexer to keep track of starting index of tokens.

lukasberglund:
You may ask, Why did I not include the length of the token. Well, I believe it to be redundant as we can just count the length of the content string. Even with escaped characters, since those will be saved using two characters in our tokens too. The tests don't work yet because they were designed without the startIndex field of the Token datatype.

Bonus change: Slightly changed the functionality of strToSymbol in order to be more elegant and haskellian. It now returns Maybe Symbol and nothing if the symbol doesn't match.


src/Token.hs
```diff
@@ -2,6 +2,9 @@ module Token where
...
 import Symbol (Symbol(..))
 
-data Token =
-  Token Symbol String
-  deriving (Eq, Show)
+
+-- startIndex - refers to the index of the first character of the token in the source code
+data Token = Token  { symbol :: Symbol
+                    , content :: String
+                    , startIndex :: Int
```
ayazhafiz: I think we should still include `end`. The reasons are as follows:

1. It's preferable to handle escaped characters in the lexer. It lends itself to a nice separation of concern, since the lexer is responsible for tokenization. It would be more inconvenient to do this in the parser, because that's much for transformation of the code into a syntax tree than evaluation of token content. Which means that the content of our tokens in the lexing stage in some cases will be different from the actual source code content.

2. Even if we don't do this, we will need `end` properties later down the line, in the parser and interpreter/tree writer, because we definitely don't want to hold onto the original content there. For consistency across modules, we should provide `start` and `end` properties on tokens as well, so that there is always one source of truth for how the indeces are defined - in the lexer. When issues come up down the line, we know immediately where to look, and changing it in one place (the lexer) will propagate the change to all downstream modules. The less transformations we have, the clearer the program is to follow.

src/Token.hs
```diff
@@ -2,6 +2,9 @@ module Token where
 
 import Symbol (Symbol(..))
 
-data Token =
-  Token Symbol String
-  deriving (Eq, Show)
+
+-- startIndex - refers to the index of the first character of the token in the source code
```
ayazhafiz: Please place this comment on top of the `startIndex` property. Also, can you rename the property to just `start`?

ayazhafiz:
Please rebase on master and fix the merge conflicts.

src/Symbol.hs
```diff
@@ -22,8 +24,6 @@ data Symbol
   | TypeDelim
   | ValueDelim
   | Assign
-  | GT
-  | LT
```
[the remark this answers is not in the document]
lukasberglund: Why?

src/Token.hs
lukasberglund: where should I place it

ayazhafiz: Just above the `start` property, as discussed offline. Marking as resolved.

ayazhafiz:
Why did you close this?

PR #14, opened 2020-01-15
Lukas

lukasberglund:
Added double keyword to language.


src/Token.hs
```diff
@@ -2,6 +2,11 @@ module Token where
 
 import Symbol (Symbol(..))
 
+
 data Token = Token  { symbol :: Symbol
                     , content :: String
+                    --refers to the index of the first character of the token in the source code
```
ayazhafiz:
```suggestion
                    -- Refers to the index of the first character of the token in the source code
```

src/Token.hs
```diff
@@ -2,6 +2,11 @@ module Token where
...
+                    , start :: Int
+                    --refers to the index of the character after the last
```
ayazhafiz:
```suggestion
                    -- Refers to the index of the character after the last
```

ayazhafiz:
@lukasberglund merge conflicts ):

PR #21, opened 2020-01-19
Added underscore functionality for names

lukasberglund:
Gave lexer the ability to lex underscores in names and wrote a test that asks for an underscore in a name. While I was at it I also wrote a test for numbers in a name.


ayazhafiz:
Closes #19.

PR #22, opened 2020-01-19
fix: lex operator combinations separately

### LINKED ISSUE #16: Lexer cannot lex things like +-1 or -(5).
It tries to evaluate all consecutive non-alphanumeric characters together, so in -(5), it tries to evaluate "-(" and returns invalid symbol.

ayazhafiz:
This commit fixes a bug that would cause the lexer to parse
possibly-ambigious operator combinations as invalid. For instance,
"-(" may be a Minus and LParen symbol, but the lexer would treat the
combination as an invalid symbol. To avoid this, the lexer now consumes
tokens incrementally and falls back on the first-valid operator
combination discovered.

Closes #16.


src/Lexer.hs
```diff
@@ -36,36 +35,97 @@ lexStr i (delim:xs) =
...
-    symbol = case keywordToSymbol keyword of
-      Just s -> s
-      _ -> Symbol.Invalid
+    (symbol, tok, remainder) = get "" str
+    end = i + length tok
+
+    get :: String -> String -> (Symbol, String, String)
+    get main [] = (Symbol.Invalid, main, [])
+    get main rem
+      | not (isAnyOf [isSymbol, isPunctuation] next) = (Symbol.Invalid, tok, rest)
```
lukasberglund: Don't you think in this case, where we have the argument,
```haskell
not (isSymbol next || isPunctuation next)
```
would be clearer?

src/Lexer.hs
```diff
@@ -36,36 +35,97 @@ lexStr i (delim:xs) =
...
-    end = i + length keyword
-    symbol = case keywordToSymbol keyword of
-      Just s -> s
-      _ -> Symbol.Invalid
+    (symbol, tok, remainder) = get "" str
+    end = i + length tok
+
+    get :: String -> String -> (Symbol, String, String)
+    get main [] = (Symbol.Invalid, main, [])
+    get main rem
```
lukasberglund: I would change this to 
```haskell
get main (next:rest)
```
and get rid of line 127

src/Lexer.hs
```diff
@@ -36,36 +35,97 @@ lexStr i (delim:xs) =
...
   where
-    (keyword, remainder) = span (isAnyOf [isSymbol, isPunctuation]) str
-    end = i + length keyword
-    symbol = case keywordToSymbol keyword of
-      Just s -> s
-      _ -> Symbol.Invalid
+    (symbol, tok, remainder) = get "" str
+    end = i + length tok
+
+    get :: String -> String -> (Symbol, String, String)
```
lukasberglund: Personally, I would make the name of the function more descriptive, "get" doesn't tell me much about what the function does.

PR #24, opened 2020-01-21
Added parser that handles math expressions, strings, variables, and function-calls.

lukasberglund:


ayazhafiz:
Hi Lukas, I think we will need to bump up the test coverage on this commit (it is rather low, at around 88%). Can you take a look at the coverage report on coveralls and add more tests to improve the coverage? Lmk if you need any help

src/Grammar.hs
```diff
@@ -0,0 +1,36 @@
+module Grammar where
+
+import Symbol
+
+
```
ayazhafiz: Extra whitespace? Can you run this through with [`hindent`](https://github.com/chrisdone/hindent)? I will add a CI test for this later on. Let me know if that sounds good or you would prefer something else.

src/Grammar.hs
```diff
@@ -0,0 +1,36 @@
...
+
+data DataType = Int8 | Int16 | Int32 | Int64 | StandardInt | UInt8 | UInt16
```
ayazhafiz: What is a `StandardInt`? Why  not just `Int`?

src/Grammar.hs
```diff
@@ -0,0 +1,36 @@
...
+                | UInt32 | UInt64 | UInt | Char
+                | Array {dataType :: DataType, size :: Int}
```
ayazhafiz: nice!

src/Grammar.hs
```diff
@@ -0,0 +1,36 @@
...
+                deriving (Show, Read, Eq)
```
ayazhafiz: to bump the test coverage, you can add tests that check that the derived versions of this act correctly

src/Grammar.hs
```diff
@@ -0,0 +1,36 @@
...
+
+data Operator = Plus | Minus | Times | Div | Mod | Negative | Positive | Paren
+                deriving (Show, Read, Eq)
+
+-- type Conditional = IfStatement | ElseStatement
+--
+-- data ElseStatement = ElseStatement
```
ayazhafiz: please remove

src/Grammar.hs
```diff
@@ -0,0 +1,36 @@
...
+
+
+type BooleanExpression = Expression
+
+-- data Sign = Positive | Negative
+
+-- data Number = Integer Int
```
ayazhafiz: please remove

src/Grammar.hs
```diff
@@ -0,0 +1,36 @@
...
+
+data Operator = Plus | Minus | Times | Div | Mod | Negative | Positive | Paren
+                deriving (Show, Read, Eq)
+
+-- type Conditional = IfStatement | ElseStatement
+--
+-- data ElseStatement = ElseStatement
+
+
+type BooleanExpression = Expression
```
ayazhafiz: so why have a separate type?

src/Grammar.hs
```diff
@@ -0,0 +1,36 @@
...
+
+-- data Number = Integer Int
+
+data Expression = InfixOperation Expression Operator Expression
+                  | PrefixOperation Operator Expression
+                  | Number Int | String String | Variable String
+                  | FunctionCall String [Expression]
+                  | Conditional {condition :: Expression
+                                  , thenArm :: Expression
+                                  , elseArm :: Expression}
```
ayazhafiz: nice!

src/Grammar.hs
```diff
@@ -0,0 +1,36 @@
...
+type BooleanExpression = Expression
+
+-- data Sign = Positive | Negative
+
+-- data Number = Integer Int
+
+data Expression = InfixOperation Expression Operator Expression
+                  | PrefixOperation Operator Expression
+                  | Number Int | String String | Variable String
+                  | FunctionCall String [Expression]
```
ayazhafiz: Can you confirm that this means a function call is a string and an array of expressions?

ayazhafiz: Would it be possible to define each of these data as separate data types? So that we have the opportunity to plug in + plug out as needed.

src/Parser.hs
```diff
@@ -0,0 +1,143 @@
+module Parser where
+
+import qualified Symbol (Symbol(..))
```
ayazhafiz: consider

```haskell
import qualified Symbol
import Symbol (Symbol(..))
```

This will allow you to use `Symbol` types with the `Symbol` namespace without having to namespace the ctor as well.

src/Parser.hs
```diff
@@ -0,0 +1,143 @@
...
+import Token (Token(..))
+import qualified Grammar
```
ayazhafiz: consider import `Expression` unqualified

src/Parser.hs
```diff
@@ -0,0 +1,143 @@
...
+
+parseExpression' :: Int -> [Token] -> (Grammar.Expression, [Token])
+parseExpression' precidence tokens = (infixExpression, afterInfix)
+  where (left, afterPrefix) = parsePrefix tokens
+        (infixExpression, afterInfix) = parseInfix precidence left afterPrefix
+
+parsePrefix :: [Token] -> (Grammar.Expression, [Token])
+-- parsePrefix - Parse operations that do not depend on previous symbols
+--               these include prefix operations like +1 or -7, but also
+--               function calls like
```
ayazhafiz: like?

src/Parser.hs
```diff
@@ -0,0 +1,143 @@
...
+symbolToPrefixOperator sym = case sym of
+  Symbol.Plus -> Grammar.Positive
+  Symbol.Minus -> Grammar.Negative
+  Symbol.LParen -> Grammar.Paren
+  Symbol.LBracket -> Grammar.Paren
+  _ -> error $ show sym
+
+operatorPrecidence :: Grammar.Operator -> Int
+operatorPrecidence op = case op of
+  Grammar.Paren -> -1
```
ayazhafiz: why negative?

test/Parser/ParserSpec.hs
```diff
@@ -0,0 +1,105 @@
...
+import Data.Semigroup
+import Prelude hiding (lex)
+import Text.Printf
+
+import Test.Hspec
+
+-- import qualified Builtin
+-- import Lexer (lex)
+-- import Symbol (Symbol(..))
+-- import Token (Token(..))
```
ayazhafiz: please remove

test/Parser/ParserSpec.hs
```diff
@@ -0,0 +1,105 @@
...
+import Parser
+import Grammar
+import Lexer
+
+spec :: Spec
+spec = do
+  describe "parseExpression" $ do
+    it "parses simple prefix expressions correctly" $ do
+      parseExpression (lex "+2") `shouldBe`
```
ayazhafiz: consider defining a helper function that will call `lex` and `parseExpression` for you.

test/Parser/ParserSpec.hs
```diff
@@ -0,0 +1,105 @@
...
+      parseExpression (lex "[ +5]") `shouldBe`
+        PrefixOperation Positive (Number 5)
+    it "rejects nonsense (1)" $ do
+      evaluate (parseExpression (lex "()")) `shouldThrow` anyException
+    it "rejects nonsense (2)" $ do
+      evaluate (parseExpression (lex "(]")) `shouldThrow` anyException
+    it "rejects nonsense (3)" $ do
+      evaluate (parseExpression (lex "(4]")) `shouldThrow` anyException
+    -- it "rejects nonsense (4)" $ do
+    --   evaluate (parseExpression (lex "5 + * 5")) `shouldThrow` anyException
```
ayazhafiz: why exclude this?

src/Grammar.hs
lukasberglund: You are right I'll just use that.

src/Grammar.hs
lukasberglund: It does.

lukasberglund: What do you mean by plug in + plug out?

src/Parser.hs
lukasberglund: I'm confused as to what this means. Are you saying that this would allow me to use things from Symbol without having to write Symbol.Whatever? If so, couldn't I just do
```haskell
import Symbol
```
to do that?

Also, I am doing import qualified because some names from Symbol clash with names from Grammar.

src/Parser.hs
lukasberglund: Because the default call to parseExpression is with a precedence of zero, but I want parseExpression to stop when it sees a closing parenthesis, hence I made it zero. If you prefer I can  shift those numbers up :)

test/Parser/ParserSpec.hs
lukasberglund: Where do you think I should define it?

test/Parser/ParserSpec.hs
lukasberglund: This one is weird. When I run it, it throws an error, but somehow it still does not not pass this test.

src/Grammar.hs
lukasberglund: Done.

src/Parser.hs
ayazhafiz: no, thats a great reason

test/Parser/ParserSpec.hs
ayazhafiz: maybe the call to `evaluate` is causing this?

lukasberglund: If I don't call evaluate it gives me this error during compilation: 
```
    • Couldn't match expected type ‘IO a0’
                  with actual type ‘Expression’
    • In the first argument of ‘shouldThrow’, namely
        ‘parseExpression (lex "5 + * 5")’
      In a stmt of a 'do' block:
        parseExpression (lex "5 + * 5") `shouldThrow` anyException
      In the second argument of ‘($)’, namely
        ‘do parseExpression (lex "5 + * 5") `shouldThrow` anyException’
   |
85 |       parseExpression (lex "5 + * 5") `shouldThrow` anyException
   |
```

lukasberglund:
> Hi Lukas, I think we will need to bump up the test coverage on this commit (it is rather low, at around 88%). Can you take a look at the coverage report on coveralls and add more tests to improve the coverage? Lmk if you need any help

I gotchu

lukasberglund:
@ayazhafiz What do you think about making a file for general helper functions like isAnyOf and splitWhen? What would you call such a file?

ayazhafiz:
Sounds great. call it Utils.

ayazhafiz:
Please rebase on master.

src/Parser.hs
```diff
@@ -0,0 +1,146 @@
...
+parseExpression :: [Token] -> Expression
+parseExpression xs = result
+  where
+    (result, []) = parseExpression' 0 xs
+
+parseExpression' :: Int -> [Token] -> (Expression, [Token])
+parseExpression' precidence tokens = (infixExpression, afterInfix)
+  where
+    (left, afterPrefix) = parsePrefix tokens
+    (infixExpression, afterInfix) = parseInfix precidence left afterPrefix
```
ayazhafiz: Please include this in the `where` clause of `parseExpression`.

src/Parser.hs
```diff
@@ -0,0 +1,146 @@
...
+
+parsePrefix :: [Token] -> (Expression, [Token])
+-- parsePrefix - Parse operations that do not depend on previous symbols.
```
ayazhafiz:
```suggestion
-- Parse operations that do not depend on previous symbols.
```

src/Parser.hs
```diff
@@ -0,0 +1,146 @@
...
+
+operatorPrecidence :: Grammar.Operator -> Int
+operatorPrecidence op =
+  case op of
+    Grammar.Paren -> -1
+    Grammar.Minus -> 1
+    Grammar.Plus -> 1
+    Grammar.Div -> 2
+    Grammar.Times -> 2
+    Grammar.Mod -> 2
```
ayazhafiz: Consider moving this to the top of the parser module.

src/Parser.hs
```diff
@@ -0,0 +1,146 @@
...
+    Symbol.LParen -> parseParen all
+    Symbol.LBracket -> parseParen all
+    Symbol.StrBound -> parseString all
+    _ -> (prefix, remainder)
+      where (expression, remainder) = parsePrefix xs
+            operator = symbolToPrefixOperator $ symbol x
+            prefix = Grammar.PrefixOperation operator expression
+
+parseInfix :: Int -> Expression -> [Token] -> (Expression, [Token])
+-- parseInfix - Given precidence and the left part of an operations parse all
```
ayazhafiz:
```suggestion
-- Given precidence and the left part of an operations parse all
```

src/Parser.hs
```diff
@@ -0,0 +1,146 @@
...
+parseEnumeration (delimToken:remainder) = (expressions, afterEnum)
+  where
+    closingDelim = getClosingDelim $ symbol delimToken
+    (insideTokens, (_:afterEnum)) = break (isSymbol closingDelim) remainder
+    splitTokens = splitWhen (isSymbol Symbol.ValueDelim) insideTokens
+    expressions = case splitTokens of
+      [[]] -> []
+      _ -> map parseExpression splitTokens
+
+getClosingDelim :: Symbol.Symbol -> Symbol.Symbol
```
ayazhafiz:
```suggestion
closingDelim :: Symbol.Symbol -> Symbol.Symbol
```

src/Parser.hs
```diff
@@ -0,0 +1,146 @@
...
+getClosingDelim s =
+  case s of
+    Symbol.LBracket -> Symbol.RBracket
+    Symbol.LParen -> Symbol.RParen
+    Symbol.StrBound -> Symbol.StrBound
+
+isSymbol :: Symbol.Symbol -> Token -> Bool
```
ayazhafiz:
```suggestion
isOfSymbol :: Symbol.Symbol -> Token -> Bool
```

src/Parser.hs
```diff
@@ -0,0 +1,146 @@
...
+symbolToOperator sym =
+  case sym of
+    Symbol.Plus -> Grammar.Plus
+    Symbol.Minus -> Grammar.Minus
+    Symbol.Times -> Grammar.Times
+    Symbol.Div -> Grammar.Div
+    Symbol.Mod -> Grammar.Mod
+    Symbol.RParen -> Grammar.Paren
+    Symbol.RBracket -> Grammar.Paren
+    _ -> error $ show sym
```
ayazhafiz: can you include a more descriptive error?

src/Parser.hs
```diff
@@ -0,0 +1,146 @@
...
+
+symbolToPrefixOperator :: Symbol.Symbol -> Grammar.Operator
+symbolToPrefixOperator sym =
+  case sym of
+    Symbol.Plus -> Grammar.Positive
+    Symbol.Minus -> Grammar.Negative
+    _ -> error $ show sym
```
ayazhafiz: ditto

test/Parser/ParserSpec.hs
```diff
@@ -0,0 +1,110 @@
...
+import Lexer
+
+spec :: Spec
+spec = do
+  describe "Parse" $ do
+    it "parses multiple expressions correctly" $ do
+      parse (lex "1 + 2! foo()!") `shouldBe`
+        [InfixOperation (Number 1) Plus (Number 2), FunctionCall "foo" []]
+    -- it "rejects empty inputs" $ do
+    --   evaluate ( parse (lex "")) `shouldThrow` anyErrorCall
```
ayazhafiz: The problem is that this doesn't throw.

test/Parser/ParserSpec.hs
ayazhafiz: At the top of this test file.

test/Parser/ParserSpec.hs
```diff
@@ -0,0 +1,110 @@
...
+      parseExpression (lex "[ +5]") `shouldBe`
+        PrefixOperation Positive (Number 5)
+    it "rejects nonsense (1)" $ do
+      evaluate (parseExpression (lex "()")) `shouldThrow` anyException
+    it "rejects nonsense (2)" $ do
+      evaluate (parseExpression (lex "(]")) `shouldThrow` anyException
+    it "rejects nonsense (3)" $ do
+      evaluate (parseExpression (lex "(4]")) `shouldThrow` anyException
+    -- it "rejects nonsense (4)" $ do
+    --   parseExpression (lex "5 + * 5") `shouldThrow` anyException
```
ayazhafiz: This doesn't throw either. It parses the expression.

test/Parser/ParserSpec.hs
```diff
@@ -0,0 +1,110 @@
...
+        InfixOperation (InfixOperation (InfixOperation (Variable "a") Div
+          (Variable "b")) Mod (Variable "c")) Plus (Variable "d")
+    it "parses parentheses correctly" $ do
+      parseExpression (lex "(3 + 5) % 2") `shouldBe`
+        InfixOperation (InfixOperation (Number 3) Plus (Number 5)) Mod (Number 2)
+      parseExpression (lex "- [ 2-2 ]") `shouldBe`
+        PrefixOperation Negative (InfixOperation (Number 2) Minus (Number 2))
+      parseExpression (lex "[ +5]") `shouldBe`
+        PrefixOperation Positive (Number 5)
+    it "rejects nonsense (1)" $ do
```
ayazhafiz: Please rename all of these tests to better reflect what they are testing, and consider grouping them in "describe" blocks. For example, all parentheses tests could go under a `describe "parentheses"` block,

src/Parser.hs
lukasberglund: I can't because I call this function from parseInfix.

test/Parser/ParserSpec.hs
lukasberglund: How do I make it throw?

src/Grammar.hs
```diff
@@ -0,0 +1,46 @@
...
+  deriving (Show, Read, Eq)
+
+data Expression
+  = InfixOperation Expression Operator Expression
+  | PrefixOperation Operator Expression
+  | Number Int
+  | String String
+  | Variable String
+  | FunctionCall String [Expression]
+  | Conditional
```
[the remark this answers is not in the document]
lukasberglund: Would that be possible given that we have to define a return type for parseExpression?

src/Parser.hs
lukasberglund: Why do you think it should go to the top of the file?

src/Grammar.hs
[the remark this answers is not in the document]
ayazhafiz: Yeah, you're right. Not sure why I thought that was possible

src/Parser.hs
ayazhafiz: fair

test/Parser/ParserSpec.hs
ayazhafiz: Change the parser so that it doesn't accept empty inputs. Or be okay with empty inputs.

test/Parser/ParserSpec.hs
```diff
@@ -0,0 +1,123 @@
...
+        PrefixOperation Positive (Number 5)
+  describe "invalid operations" $ do
+    it "rejects empty parentheses" $ do
+      evaluate (parseExpression (lex "()")) `shouldThrow` anyException
+    it "rejects parentheses that don't match (1)" $ do
+      evaluate (parseExpression (lex "(]")) `shouldThrow` anyException
+    it "rejects parentheses that don't match (2)" $ do
+      evaluate (parseExpression (lex "(4]")) `shouldThrow` anyException
+    -- it "rejects two infix operators following one another" $ do
+    --   parseExpression (lex "5 + * 5") `shouldThrow` anyException
```
ayazhafiz: can you add code so that the parser actually throws in this case?

lukasberglund: I don't know how to make it actually throw
```haskell
error "some message"
```
doesn't  work apparently.

ayazhafiz:
That's weird. Can you show me the code you are using?

test/Parser/ParserSpec.hs
lukasberglund: This is another case of me throwing an error but it not really recognizing that I did throw an error.

src/Parser.hs
```diff
@@ -0,0 +1,139 @@
...
+closingDelim :: Symbol -> Symbol
+closingDelim s =
+  case s of
+    Symbol.LBracket -> Symbol.RBracket
+    Symbol.LParen -> Symbol.RParen
+
+isOfSymbol :: Symbol -> Token -> Bool
+isOfSymbol sym token = (symbol token == sym)
+
+parse :: [Token] -> [Expression]
```
lukasberglund: @ayazhafiz Would you want this to return something like [(Expression, Error)]?

src/Parser.hs
```diff
@@ -0,0 +1,139 @@
...
+getLines tokens = (firstLine:nextLines)
+    where (firstLine, (exprEnd:remainder)) = break (isOfSymbol Symbol.ExprEnd) tokens
+          nextLines = getLines remainder
+
+parseExpression :: [Token] -> Expression
+parseExpression xs = result
+  where
+    (result, []) = parseExpression' 0 xs
+
+parseExpression' :: Int -> [Token] -> (Expression, [Token])
```
lukasberglund: @ayazhafiz If I have to add an error here, what would you suggest I make this function return?

lukasberglund: On second thought I would probably just make it return (Expression, Error, [Token])

src/Parser.hs
ayazhafiz: That sounds good to me. Maybe more like `([Expression], [Error])`.

ayazhafiz:
How is it going with the errors? Can I help?

lukasberglund:
> How is it going with the errors? Can I help?

Yes! I started with something. It runs but it doesn't work for all types of errors yet and I haven't rewritten the tests to reflect the changes I made to the parser. 

It would be great if you told me if I am on the wrong track or whether I should change something about my approach. Getting hrer was surprisingly difficult!

ayazhafiz:
Okay, cool. I will take a look soon!

lukasberglund:
> For example, an InfixOperation should always be an `Expression Operator Expression`, not `(Maybe Expression) Operator (Maybe Expression)` (in the latter case, if the LHS was a Nothing, the whole expression wouldn't be an infix operation)! Something I would do is to try to record as much as possible in my AST before bailing out. For example, `1.2.3 + 5` could parsed something like `Number [1.] InfixOperator [Number [2.3] Plus Number [5]]` or `InfixOperator[Unknown [1.2.3] Plus Number [5]]`. This way, we preserve the structure of the grammar without having to have `Maybe`s, and can keep the errors independent of the AST.

Alright, I changed it up as per your recommendation. The only thing that is annoying now it that I created a seperate type of Blank datatype for each grammar structure (BlankOperator, BlankExpression, etc.) which doesn't seem very elegant.

> There are probably even better ways to approach this. Feel free to explore and implement how you see best or most interesting!
 :)

ayazhafiz:
> The only thing that is annoying now it that I created a seperate type of Blank datatype for each grammar structure (BlankOperator, BlankExpression, etc.) which doesn't seem very elegant.

Yeah, this is probably not the way we should go. I would recommend looking up how other parsers handle this (for example, check out Rust and TypeScript -- if you need help finding the sources, let me know and I'll point you to the relevant code). In cases where you need a blank operator or expression, we probably just shouldn't recover at all. It can be a best-effort kind of thing. But yeah something like `1 $$ 4` isn't parsable at all after `1 Invalid[$$] 4`.

lukasberglund:
> Yeah, this is probably not the way we should go. I would recommend looking up how other parsers handle this (for example, check out Rust and TypeScript -- if you need help finding the sources, let me know and I'll point you to the relevant code). In cases where you need a blank operator or expression, we probably just shouldn't recover at all. It can be a best-effort kind of thing. But yeah something like `1 $$ 4` isn't parsable at all after `1 Invalid[$$] 4`.

Could you point me to the relevant code in rust? I'm having a hard time grasping what's going on.

ayazhafiz:
Sure, I got you. Give me some time.

ayazhafiz:
@lukasberglund https://github.com/rust-lang/rustc-guide/blob/master/src/the-parser.md

lukasberglund:
I am having trouble understanding what Rust does, partially due to the fact I have never written in Rust. I am going to try to see what tutorials say on how to handle errors in parsers. Lmk if you think that is a bad idea or if there are tutorials you would recommend.

ayazhafiz:
Seems fine to me. Not really aware of any tutorials.

Отправлено: Saturday, February 22, 2020 11:47:10 PM
Копия: hafiz <ayaz.hafiz.1@gmail.com>; Mention <mention@noreply.github.com>
Тема: Re: [ayazhafiz/hgb] Added parser that handles math expressions, strings, variables, and function-calls. (#24)


I am having trouble understanding what Rust does, partially due to the fact I have never written in Rust. I am going to try to see what tutorials say on how to handle errors in parsers. Lmk if you think that is a bad idea or if there are tutorials you would recommend.

lukasberglund:
Okay so I changed things up so that now the parser gives up when it hits any parse errors. I think this makes sense since it makes the code a lot more readible and a lot of compilers seem to do it that way.

test/Lexer/LexerSpec.hs
```diff
@@ -13,26 +13,32 @@ import Lexer (lex)
 import Symbol (Symbol(..))
 import Token (Token(..))
 
+lexRemoveEndOfFile :: String -> [Token]
```
ayazhafiz: just call this `driveLexer` or something and leave a comment that it removes EOF

ayazhafiz: Bump for this

lukasberglund:
Hey mate. Can you take a look at coveralls and the tests? Ive tried a bunch of things but I can't get the coverage up

ayazhafiz:
Sure

Отправлено: Wednesday, March 4, 2020 9:43:36 PM
Копия: hafiz <ayaz.hafiz.1@gmail.com>; Mention <mention@noreply.github.com>
Тема: Re: [ayazhafiz/hgb] Added parser that handles math expressions, strings, variables, and function-calls. (#24)


Hey mate. Can you take a look at coveralls and the tests? Ive tried a bunch of things but I can't get the coverage up

src/Grammar.hs
```diff
@@ -0,0 +1,47 @@
...
+  | Negative
+  | Positive
+  | Paren
+  deriving (Show, Eq)
+
+data Expression
+  = Operation
+      { name :: String
+      , arguments :: [Expression]
+      }
```
ayazhafiz: Is this supposed to represent a function-like operation (func [param1, param2]), a binary-like operation (lessThan [2, 3]), both, or something else? I think we can be more specific here.

test/Parser/ParserSpec.hs
```diff
@@ -0,0 +1,171 @@
...
+          [ Operation
+              "Mod"
+              [Operation "Div" [Variable "a", Variable "b"], Variable "c"]
+          , Variable "d"
+          ]
+    it "deals with parentheses correctly" $ do
+      parseLineNoRemainder "(3 + 5) % 2!" `shouldBe`
+        Operation "Mod" [Operation "Plus" [Number 3, (Number 5)], Number 2]
+      parseLineNoRemainder "- [ 2-2 ]!" `shouldBe`
+        Operation "Minus" [Operation "Minus" [(Number 2), (Number 2)]]
```
ayazhafiz: Looks like we are recording Operators as operations rather via the Operator grammar. Why is this? Should Operation be updated to take in Operator as the name? In that case, we should also distinguish between e.g. BinaryOperations as in most of these cases...

ayazhafiz:
Actually, weird that coveralls didn't pick up the new coverage changes too much... on my machine it's up to 98% now. I will investigate later, have to go to work rn...

src/Error.hs
```diff
@@ -5,7 +5,7 @@ import Token (Token(..))
 
 data Error =
   Error ErrorType Token
-  deriving (Show, Read, Eq)
```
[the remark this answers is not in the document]
ayazhafiz: Why derive it if we don't need it? That is part of what coverage helps us catch.

src/Grammar.hs
ayazhafiz: Hmm.. so do we permit pre and postfix function operations with no differentiation? E.g. you are suggesting 

```
func op1 op2
```

and

```
op1 + op2
```

are represented the same way internally (`Operation "func" [op1, op2]` and `Operation "+" [op1, op2]`).

This is a powerful pattern, but there are a couple of limitations:

- Complicates the parser (have multiple rules for the same output of "Operation" expression)
- Defers validation of `+` function to a later compiler stage. But at the parser level we have enough information to know that `+` can only ever have two operands, in an infix syntax.

What do you think?

test/Parser/ParserSpec.hs
ayazhafiz: Why not

```haskell
Operation
   { name :: Operator
    , arguments :: [Expression]
    }
```

ayazhafiz: Well, unless we decide that this grammar (`Operation`) should only be for functions. Then we should probably keep as a string.

src/Error.hs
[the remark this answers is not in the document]
lukasberglund: True, I just figured we are making our datatype less useful than it could be. But I guess we can always add it later.

src/Grammar.hs
lukasberglund: > Hmm.. so do we permit pre and postfix function operations with no differentiation? E.g. you are suggesting func op1 op2 and op1 + op2 are represented the same way internally (Operation "func" [op1, op2] and Operation "+" [op1, op2]).

Yes I am suggesting they are represented the same way internally, although 
`func op1 op2`
would currently have to be written as 
`func(op1 op2)`

> Complicates the parser (have multiple rules for the same output of "Operation" expression)

Fair enough, if you want to, I can add InfixOperation and PrefixOperation again.

> Defers validation of + function to a later compiler stage. But at the parser level we have enough information to know that + can only ever have two operands, in an infix syntax.

Well the parser is currently written so that it will only parse infix operation that have to operands, so we are performing validation.

lukasberglund: Lmk if I should add prefix and infix operations.

ayazhafiz: How much do you think it will complicate our parser if we choose to use only one Operation?

Some other things to think about:

- How will we evaluate/transform these operations? For binary infix operations do we just assume a contract that there are only two parameters? This could lead to some assumptions in the state, whereas in a binary operation AST we can guarantee the number of parameters

- How will we express the name of an operation? User defined functions will have to be a string. Should built in infix operations be string defined too?

lukasberglund: > How much do you think it will complicate our parser if we choose to use only one Operation?

I'm not sure I fully understand what you are asking. The parser is currently written to use only one operation. 

> How will we evaluate/transform these operations? For binary infix operations do we just assume a contract that there are only two parameters? This could lead to some assumptions in the state, whereas in a binary operation AST we can guarantee the number of parameters

Well it's true that we can't know for sure that there are two parameters in a sense (although the parser would never produce a binary operation that doesn't have more than two parameters). But isn't it the same for function calls? If the user defines a function that takes two parameters, we can't be sure that when he calls the function he will call it with two parameters.

> How will we express the name of an operation? User defined functions will have to be a string. Should built in infix operations be string defined too?

I think the best way to do this might be to edit the operation constructor to something like.
```haskell
Operation {opType :: OperationType, arguments :: [Expression]}
```
and define OperationType a bit like this
```haskell
data OperationType = Plus | Minus | ... | UserFunction String
```

ayazhafiz:
Okay, sounds good to me. Let’s try it out! Can you fix with changes regarding OperationType?

Another option regarding OperationType is just keep the name as a string. Then we can an expression like 1+2 parsed as an operation with name “+”. IMO this might be better so that we can perform lookup in a function table in a consist way later on.

Отправлено: Friday, March 6, 2020 6:50:31 PM
Копия: hafiz <ayaz.hafiz.1@gmail.com>; Mention <mention@noreply.github.com>
Тема: Re: [ayazhafiz/hgb] Added parser that handles math expressions, strings, variables, and function-calls. (#24)


@lukasberglund commented on this pull request.

In src/Grammar.hs<https://github.com/ayazhafiz/hgb/pull/24#discussion_r389221035>:

> +  = Operation
+      { name :: String
+      , arguments :: [Expression]
+      }


How much do you think it will complicate our parser if we choose to use only one Operation?

I'm not sure I fully understand what you are asking. The parser is currently written to use only one operation.

How will we evaluate/transform these operations? For binary infix operations do we just assume a contract that there are only two parameters? This could lead to some assumptions in the state, whereas in a binary operation AST we can guarantee the number of parameters

Well it's true that we can't know for sure that there are two parameters in a sense (although the parser would never produce a binary operation that doesn't have more than two parameters). But isn't it the same for function calls? If the user defines a function that takes two parameters, we can't be sure that when he calls the function he will call it with two parameters.

How will we express the name of an operation? User defined functions will have to be a string. Should built in infix operations be string defined too?

I think the best way to do this might be to edit the operation constructor to something like.

Operation {opType :: OperationType, arguments :: [Expression]}

and define OperationType a bit like this

data OperationType = Plus | Minus | ... | UserFunction String

ayazhafiz:
Okay, sounds good to me. Let’s try it out! Can you fix with changes regarding OperationType?

Another option regarding OperationType is just keep the name as a string. Then we can an expression like 1+2 parsed as an operation with name “+”. IMO this might be better so that we can perform lookup in a function table in a consist way later on.

Отправлено: Friday, March 6, 2020 6:50:31 PM
Копия: hafiz <ayaz.hafiz.1@gmail.com>; Mention <mention@noreply.github.com>
Тема: Re: [ayazhafiz/hgb] Added parser that handles math expressions, strings, variables, and function-calls. (#24)


@lukasberglund commented on this pull request.

In src/Grammar.hs<https://github.com/ayazhafiz/hgb/pull/24#discussion_r389221035>:

> +  = Operation+      { name :: String+      , arguments :: [Expression]+      }

How much do you think it will complicate our parser if we choose to use only one Operation?

I'm not sure I fully understand what you are asking. The parser is currently written to use only one operation.

How will we evaluate/transform these operations? For binary infix operations do we just assume a contract that there are only two parameters? This could lead to some assumptions in the state, whereas in a binary operation AST we can guarantee the number of parameters

Well it's true that we can't know for sure that there are two parameters in a sense (although the parser would never produce a binary operation that doesn't have more than two parameters). But isn't it the same for function calls? If the user defines a function that takes two parameters, we can't be sure that when he calls the function he will call it with two parameters.

How will we express the name of an operation? User defined functions will have to be a string. Should built in infix operations be string defined too?

I think the best way to do this might be to edit the operation constructor to something like.

Operation {opType :: OperationType, arguments :: [Expression]}

and define OperationType a bit like this

data OperationType = Plus | Minus | ... | UserFunction String

lukasberglund:
> Okay, sounds good to me. Let’s try it out! Can you fix with changes regarding OperationType? Another option regarding OperationType is just keep the name as a string. Then we can an expression like 1+2 parsed as an operation with name “+”. IMO this might be better so that we can perform lookup in a function table in a consist way later on.

I did what you suggested as an alternative option!

src/Grammar.hs
```diff
@@ -0,0 +1,47 @@
...
+data Operator
+  = Plus
+  | Minus
+  | Times
+  | Div
+  | Mod
+  | Negative
+  | Positive
+  | Paren
+  deriving (Show, Eq)
```
ayazhafiz: Will you get rid of this since we are moving to all functions having string names?

lukasberglund:
It looks like all that's left is to bump up the test coverage. Could you take a look at that?

ayazhafiz:
Sure

ayazhafiz:
Hey Lukas, looks like the git history for this branch is pretty messed up (with branching). This makes rebasing pretty difficult. In the future let's take care to keep history on one branch so we don't have to merge branches and can rebase instead.

ayazhafiz:
@lukasberglund looks like the error to catch missing expression endings is not working correctly:

```
    it "should expect ! to end expressions" $ do
      parseError "1+1" `shouldBeError` ExpectedSymbol Symbol.ExprEnd

--- 

  1) Parser.Parser.errors should expect ! to end expressions
       expected: ExpectedSymbol ExprEnd
        but got: ExpectedOperator
```

Do you want to fix this? Or should I give it a shot

lukasberglund:
> Hey Lukas, looks like the git history for this branch is pretty messed up (with branching). This makes rebasing pretty difficult. In the future let's take care to keep history on one branch so we don't have to merge branches and can rebase instead.

Sorry about that. I'm not quite sure what it means to keep history on one branch. What should I do/ not do to make sure that happens?

Also, I'll fix the error.

lukasberglund:
Okay I've simplified the errors let me know what you think. I changed the code of the parser to take advantage of the fact that Either is a monad (I just learned about Monads!) I wanted to put that in a later commit but somehow it snuck it's way into this one. I'm gonna make it a bit more readible in the next half hour so maybe read the code then.

ayazhafiz:
Yay!

Отправлено: Saturday, March 7, 2020 9:53:13 PM
Копия: hafiz <ayaz.hafiz.1@gmail.com>; Mention <mention@noreply.github.com>
Тема: Re: [ayazhafiz/hgb] Added parser that handles math expressions, strings, variables, and function-calls. (#24)


Okay I've simplified the errors let me know what you think. I changed the code of the parser to take advantage of the fact that Either is a monad (I just learned about Monads!) I wanted to put that in a later commit but somehow it snuck it's way into this one. I'm gonna make it a bit more readible in the next half hour so maybe read the code then.

lukasberglund:
@ayazhafiz  so right now the ordering of our functions is a bit weird. First we have simple mapping functions then we go from high-level to low level. Would it be more true to convention and better to just go from most low level at the top to most high level at the bottom?

ayazhafiz:
> @ayazhafiz so right now the ordering of our functions is a bit weird. First we have simple mapping functions then we go from high-level to low level. Would it be more true to convention and better to just go from most low level at the top to most high level at the bottom?

Yes

src/Parser.hs
```diff
@@ -0,0 +1,176 @@
+{-# LANGUAGE MultiWayIf #-}
```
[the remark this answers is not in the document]
lukasberglund: Not sure how it works but it allows for the use of guards (these guys --> '|') in a do structure through this syntax:

```haskell
do
...
if
 | a == b -> 1
 | otherwise -> 2
```
or something like that

src/Parser.hs
```diff
@@ -0,0 +1,176 @@
...
+-- 1. Parse prefix operators, which in this context means any operator that does
+--    not rely on context from before (including negative/positive signs but also
+--    function calls and opening parentheses)
+-- 2. Parse infix operators, which in this case are operators that rely on context
+--    from before. Here, this means operators like '+', '-', and '=='. Note that
+--    parseInfix deals with precidence and commutativity by using a precidence
+--    operator that was passed in as an argument. That means, if it encounters
+--    an operator that is lower precidence than what it is parsing it stops.
+--    As I am writing this even I am not quite sure why this works but somehow
+--    through the power of recursion everything turns out right.
```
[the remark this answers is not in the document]
lukasberglund: That's a pretty good explanation, thanks!

lukasberglund:
> > @ayazhafiz so right now the ordering of our functions is a bit weird. First we have simple mapping functions then we go from high-level to low level. Would it be more true to convention and better to just go from most low level at the top to most high level at the bottom?
> 
> Yes

I'll do that then.

src/Parser.hs
```diff
@@ -0,0 +1,176 @@
...
+  let (x:xs) = afterExpression
+  if | isOfSymbol Symbol.ExprEnd x -> (expression :) <$> parse xs
+     | otherwise -> Left $ Error (Expected $ show Symbol.ExprEnd) x
+
+parseExpression :: [Token] -> Either Error (Expression, [Token])
+parseExpression = parseExpression' 0
+
+parseExpression' :: Int -> [Token] -> Either Error (Expression, [Token])
+parseExpression' precidence tokens = do
+  (left, afterLeft) <- parsePrefix tokens
```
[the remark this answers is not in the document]
lukasberglund: A do statement allows simplified syntax for monadic structures. 

`Either a b` is a monad with the value that's wrapped in the monad being whatever is on the right (In this case that's `(Expression, [Token])`).  The do statement allows me to  use statements like 
```haskell
(left, afterLeft) <- parsePrefix tokens
```
to extract the values out of monads without having to deal with failure (in this case an `Error` value).  In this example if `parsePrefix tokens` returns an `Right (Expression, [Token])` value as expected it gets assigned to `(left, afterLeft)`. If `parsePrefix tokens` returns an error then parseExpression' stops and just returns that error.

An alternative way of writing  this would be 
```haskell
parseExpression' precidence tokens = 
    parsePrefix tokens >>=
        (left, afterLeft) -> ...
```
Or if you haven't learned about monads yet, you could just write
```haskell
parseExpression' precidence tokens = 
    case parsePrefix tokens of
        Left e -> Left e
        Right (left, afterLeft) -> ...
```

src/Parser.hs
```diff
@@ -0,0 +1,176 @@
...
+parseInfix :: Int -> Expression -> [Token] -> Either Error (Expression, [Token])
+-- Given precidence and the left part of an operations parse all operations to
+-- the right with a higher precidence than the one called with.
+-- The idea: Check what the precidence of the next operator is, if it is smaller than
+-- or equal to the precidence we have inputted then we stop, otherwise we parse
+-- an infix expression
+parseInfix precidence left tokens@(operatorTok:afterOperator)
+  | terminatesExpressions operatorSym = Right (left, tokens)
+  | not (isInfixOperator operatorSym) =
+    Left $ Error (Expected "Operator or '!'") operatorTok
```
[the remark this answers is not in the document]
lukasberglund: Hindent actually did this.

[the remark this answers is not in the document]
ayazhafiz: Sounds good

src/Parser.hs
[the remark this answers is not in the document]
ayazhafiz: Interesting. I like it. Much pattern match.

src/Parser.hs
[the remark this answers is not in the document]
ayazhafiz: In Haskell, do all monads have a `Left` and `Right`? Or is this that just how `Error` implements a monad?

Is it correct that `<-` is just syntatic sugar for `>>=` with the additional benefit that it bring `left` and `afterLeft` into scope as variables?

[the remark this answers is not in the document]
lukasberglund: It might help to skim [learnyouahaskell.com/a-fistful-of-monads](url) to see how monads work in haskell. 

> In Haskell, do all monads have a Left and Right? Or is this that just how Error implements a monad?

Not quite. In Haskell, `Either a b` is a monad, but `IO` and `Maybe`, and even lists are monads too.  For example if `parseExpression'` and `parsePrefix` returned the monad `Maybe (Expression, [Token])`, then 
```haskell
(left, afterLeft) <- parsePrefix tokens
```
would do the same thing if `parsePrefix tokens` returned `Just (Expression, [Token])` but if `parsePrefix tokens` returned `Nothing` then all of `parseExpression`  returns `Nothing`. 

> Is it correct that <- is just syntatic sugar for >>= with the additional benefit that it bring left and afterLeft into scope as variables?

Yes, it's just syntactic sugar for `>>=` but there is not additional benefit. `>>=` is an infix operator that takes a monad and a function that returns a monad and returns a monad so it's declaration is something like.
```haskell
(>>=) :: m a -> (a -> m b) -> m b
```
 Thus by writing

```haskell
parseExpression' precidence tokens = 
    parsePrefix tokens >>=
        \(left, afterLeft) -> ...
```
You can finish writing what's behind the error as the rest of your function definition with `(left, afterLeft)` in scope. However this is kind of awkward because you have to define a lambda inside of your function.

PR #27, opened 2020-03-07
Improved code on lexer.

lukasberglund:
Made a few improvements to the lexer's code. The main change is making the helper function return a tuple of the lexed token and the remainder so that the lexer can then call itself on the remainder. Before, the helper functions would directly call lex on the remainder. I think this is an improvement because now the helper functions do what their name implies (they only lex a number or a string or whatever instead of lexing everything). 

The change to Token.hs was a result of me running hindent.


src/Lexer.hs
```diff
@@ -15,51 +15,52 @@ lex' :: Int -> String -> [Token]
...
-          Token Symbol.StrBound [endDelim] end (end + 1) :
-          lex' (end + 1) remainder
-        _ -> []
+    (contentStr, afterContent) = break (== delim) afterOpener
+    contentStart = i + 1
+    contentEnd = contentStart + length contentStr
+    openingDelimToken = Token Symbol.StrBound [delim] i contentStart
+    contentToken = Token Symbol.String contentStr contentStart contentEnd
+    closingDelimToken =
+      Token Symbol.StrBound [delim] contentEnd (contentEnd + 1)
```
ayazhafiz: nice!

src/Token.hs
```diff
@@ -2,11 +2,13 @@ module Token where
...
 import Symbol (Symbol(..))
 
-
-data Token = Token  { symbol :: Symbol
-                    , content :: String
+data Token =
+  Token
+    { symbol :: Symbol
+    , content :: String
                     -- Refers to the index of the first character of the token in the source code
```
ayazhafiz: Let's align the comments

PR #28, opened 2020-03-15
Change parser test strategy

ayazhafiz:
This branch presents a strategy for testing hgb's parser in a humanized
manner by comparing expected parse trees and error messages to
human-readable representations rather than data-based representations.
This provides for a succinct testing strategy that is easy to extend.

Parse trees are compared using a Lisp-like intermediate representation.
The representation effectively validates that the parse tree is in the
expected hierarchy via a series a infix operators.


ayazhafiz:
@lukasberglund this is still WIP (I will finish up tomorrow), but PTAL and let me know if you think this is a good track to take. I think it will make it easier to add tests; I am doing this to see how we can boost coverage. On my machine, we are currently at

Screen Shot 2020-03-14 at 9 45 28 PM

Not totally sure what "Alternatives" are, but I am hoping this will give more insight as I am able to grow a large test corpus.

src/Grammar.hs
```diff
@@ -1,36 +1,18 @@
...
-  | UInt8
-  | UInt16
-  | UInt32
-  | UInt64
-  | UInt
-  | Char
-  | Array
-      { dataType :: DataType
-      , size :: Int
-      }
```
ayazhafiz: I have removed this for now because we do not currently support arrays. Same for `Expression.Conditional`.

test/Parser/ParserSpec.hs
```diff
@@ -16,6 +16,13 @@ import qualified Symbol
 import Test.Hspec
 import Token
 
+-- Parses a into shown expressions.
+-- See the Show instances in src/Grammar.hs for the expected format.
+driveParser :: String -> [String]
+driveParser s = map show expressions
+  where
+    Right expressions = (parse . lex) s
```
lukasberglund:
```suggestion
driveParser :: String -> Maybe Error [String]
driveParser s = map show <$> (parse . lex) s
```
When I was writing the parser I ran into the problem where I would parse a correct expression but since my parser was faulty it parsed it as an Error. However I couldn't see what kind of error it parsed it as because the pattern match `Right expression` expected the parsing to succeed. What ended up happening is that the test program would give an error because of the failed pattern match. In order to see what the error was I would then have to add a line of code to the test program that parsed the same expression again but allowed for errors , which was a hassle.  

Therefore I reckon that having driveParser return either error or expression will make debugging a lot easier.

Also if we want to simplify tests further, we could have the parser just return a string that represents the whole result of the parsing operation. I.e we could write 
```haskell
driveParser :: String -> String
driveParser s = show (parse . lex) s
```
That way we wouldn't have to use different statements when we expect an `Error` vs when we expect an `[Expression]`

src/Error.hs
```diff
@@ -3,10 +3,16 @@ module Error where
...
 data Error =
   Error ErrorType Token
-  deriving (Show, Eq)
+  deriving (Eq)
 
-data ErrorType =
-  Expected String
-  deriving (Show, Eq)
+instance Show Error where
+  show (Error errorType tok) = show errorType ++ "; got " ++ show tok
```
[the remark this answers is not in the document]
ayazhafiz: Done.

src/Token.hs
```diff
@@ -11,4 +11,7 @@ data Token =
     -- Refers to the index of the character after the last
     , end :: Int
     }
-  deriving (Eq, Show)
+  deriving (Eq)
+
+instance Show Token where
+  show (Token _ content start end) = content ++ "@" ++ show start ++ ":" ++ show end
```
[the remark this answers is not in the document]
ayazhafiz: I have updated this; PTAL

src/Grammar.hs
```diff
@@ -1,36 +1,18 @@
...
   = Number Int
   | String String
   | Variable String
-  | Conditional
-      { condition :: Expression
-      , thenArm :: Expression
-      , elseArm :: Expression
-      }
-  | Operation
+  | Call
```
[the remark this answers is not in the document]
ayazhafiz: Yep, will do

test/Parser/ParserSpec.hs
ayazhafiz: Great idea. Done 👍

ayazhafiz:
@lukasberglund PTAL updated. I have also started adding a lot of comments.. let me know if it is too much.

src/Lexer.hs
```diff
@@ -93,56 +95,49 @@ lexAlphaKeyword i str = (Token symbol keyword i end, afterKeyword)
...
         tok = main ++ [next]
+    backupInvalid ::
+         (String -> String -> SymbolWithContent)
+      -> SymbolWithContent
+      -> SymbolWithContent
+    backupInvalid _ backup@(_, _, []) = backup
+    backupInvalid getSymbol backup@(_, contentToken, contentRest) =
+      case getSymbol contentToken contentRest of
+        (Symbol.Invalid, _, _) -> backup
+        token -> token
```
lukasberglund: This is impressive. Nice.

src/Error.hs
```diff
@@ -1,18 +1,21 @@
...
 instance Show ErrorType where
-  show (Expected s) = "Expected " ++ s
+  show (ExpectedSymbol expected actual) = "expected " ++ show expected  ++ "; got " ++ show actual
 
-data Error =
-  Error ErrorType Token
-  deriving (Eq)
+newtype Error =
+  -- TODO(ayazhafiz): include the location of an error
+  Error ErrorType
```
lukasberglund: Looking at the rest of our code I feel like at this point we may have gone full circle and we should just define error like this
```haskell
data Error = ExpectedSymbol 
    {
      expected :: Any,
      actual :: One
    }
```
With the options for creating lexer errors and other errors later on. It seems like right now the error constructor does not add anything to the errortype type.

src/Parser.hs
```diff
@@ -87,15 +62,27 @@ parseExpression' precidence tokens = do
...
+            [Symbol.LParen, Symbol.Number, Symbol.Name, Symbol.StrBound]
+        , actual = One sym
+        }
   where
-    s = symbol x
+    prefixOperators = [Symbol.Plus, Symbol.Minus]
+    sym = symbol tok
+
+wrapError :: ErrorType -> Either Error any
+wrapError err = Left (Error err)
```
lukasberglund: Before I read the definition I found `wrapError` kind of confusing. I think if we change `Error` to be identical to `ErrorType` as I suggested earlier the need for `wrapError` will go away and our code will become easier to read.
```suggestion
```

src/Parser.hs
```diff
@@ -104,67 +91,119 @@ parseInfix :: Int -> Expression -> [Token] -> Either Error (Expression, [Token])
...
+  (arguments, afterArguments) <-
+    parseParameterList listOpen parameterListAndBeyond
   return (Grammar.Call (content name) arguments, afterArguments)
 
-parseEnumeration :: [Token] -> Either Error ([Expression], [Token])
-parseEnumeration (opener:afterOpener@(x:xs))
-  | isOfSymbol closer x = Right ([], xs)
-  | otherwise = parseEnumeration' afterOpener
+parseParameterList :: Token -> [Token] -> Either Error ([Expression], [Token])
+parseParameterList
```
lukasberglund: I think parseEnumeration is a better name since I want to reuse this later for array declarations.

src/Parser.hs
```diff
@@ -104,67 +91,119 @@ parseInfix :: Int -> Expression -> [Token] -> Either Error (Expression, [Token])
...
-              return (expression : otherExpressions, afterEnumeration)
-         | otherwise -> Left $ Error (Expected "Enumerator") x
+    listClose = closingDelim $ symbol listOpen
+    parseParameters' :: [Token] -> Either Error ([Expression], [Token])
+    parseParameters' params = do
+      (parameter, afterParameter) <- parseExpression params -- parse the first parameter
+      -- Parse tokens after the parameter. There are three cases here:
+      --   1. the parameter is followed by a parameter delimiter
+      --   2. the parameter is followed by a list close
+      --   3. neither (1) nor (2)
```
lukasberglund:
```suggestion
      --   3. neither (1) nor (2), which means the syntax is invalid
```

src/Parser.hs
```diff
@@ -104,67 +91,119 @@ parseInfix :: Int -> Expression -> [Token] -> Either Error (Expression, [Token])
...
+           do (otherParameters, beyondList) <-
+                parseParameters' restOfListAndBeyond
+              return (parameter : otherParameters, beyondList)
+         | sym == listClose
+          -- 2. The parameter is followed by a list close
+          --    param)!
+          --         ^  | list close
+          --          ^ | rest of list and beyond
+          --    In this case, just return the one parameter.
+          -> Right ([parameter], restOfListAndBeyond)
```
lukasberglund:
```suggestion
          -> return ([parameter], restOfListAndBeyond)
```

test/Parser/ParserSpec.hs
```diff
@@ -2,84 +2,42 @@ module Parser.ParserSpec
...
-    Left e = (parse . lex) s
-
-shouldError :: Error -> ErrorType -> Expectation
-shouldError x = shouldBe t
-  where
-    Error t _ = x
+-- Parses a Gingerbread program into Shown expressions or errors.
+-- See the Show instances in src/Grammar.hs and src/Error.hs for expected formats.
+driveParser :: String -> String
+driveParser s = lex s & parse & either show (unwords . map show)
```
lukasberglund: Fucking beautiful.

src/Symbol.hs
```diff
@@ -46,4 +49,37 @@ data Symbol
...
+    where
+      symbols' = map One symbols
+      concat :: [One] -> String
+      concat lst =
+        case lst
+          -- ..., AnotherOption, or YetAnotherOption
+              of
+          [slast, last] -> show slast ++ ", or " ++ show last
+          -- Option, ...
+          (first:rest) -> show first ++ ", " ++ concat rest
```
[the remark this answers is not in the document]
lukasberglund: You could have spared quite some work if you didn't use an Oxford comma lol.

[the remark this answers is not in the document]
ayazhafiz: lol. true!

src/Symbol.hs
```diff
@@ -46,4 +49,37 @@ data Symbol
   | Number
   | Name
   | EndOfFile
+  -- TODO(ayazhafiz): it would be nicer if we showed the expected symbol token rather than the
+  -- symbol name.
   deriving (Eq, Show)
+
+-- Symbol options
+newtype One =
```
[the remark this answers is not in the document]
ayazhafiz: Yeah, I guess the point here is just to make things a little bit more easy to construct -- now we can display the options just by calling `show` recursively on data members.
I don't disagree that the names are opaque. There are a couple options -- rename to something like `OnlySymbol` and `AnySymbol`, or add documentation, or both.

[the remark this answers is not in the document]
ayazhafiz: What do you think?

src/Error.hs
ayazhafiz: I actually like the idea of seperating an error from an error type. This means we can keep some generic data about the error, like its location (which we should add later on) from construction of the error itself. Later on I could see the need for a variety of different error constructions (e.g. a bad function call error), but how we define the location might remain the same for all. In this case it is better to not duplicate the location and definition and we get a guarantee that location is on every error message. What do you think?

src/Parser.hs
ayazhafiz: Dependent on the above comment

src/Parser.hs
ayazhafiz: Good call, will change

src/Parser.hs
ayazhafiz: Done

ayazhafiz:
Screen Shot 2020-03-15 at 11 49 58 PM

Spent an hour trying to figure out what the missing alternatives in the Grammar were, but...  🤤

src/Error.hs
lukasberglund: Good point.

src/Parser.hs
```diff
@@ -81,24 +80,65 @@ parsePrefix tokens@(tok:_)
     prefixOperators = [Symbol.Plus, Symbol.Minus]
     sym = symbol tok
 
-wrapError :: ErrorType -> Either Error any
-wrapError err = Left (Error err)
-
+-- Parse expressions that depend on previous parsing contexts.
```
lukasberglund: Can we somehow make it clear that these comments refer to `parseInfix` which is defined below?

src/Symbol.hs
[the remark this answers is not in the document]
lukasberglund: I've actually warmed up to the idea of using `One` and `Any`. I think they are useful for errors and I think we can make them generic so they don't have to be about `Symbol` by doing something like this:
```haskell
data (show a) One a = One a

instance Show One a where
   show (One a) = typeOf a ++ " " ++ show a
```
To use `typeOf` I think we have to import `Data.Typeable`. If we do this we should put it in the Error.hs file.

src/Parser.hs
ayazhafiz: Is it not clarified by their being above the function definition?

ayazhafiz:
> Actually forget all that. I thought about it some more and I think I have a way to simplify all this and make the code clearer. I reckon we should change ExpectedSymbol to take [Symbol] as expected and Symbol as got.

Great idea. Fixed; PTAL.

ayazhafiz:
Yes

Отправлено: Monday, March 16, 2020 2:25:31 PM
Копия: hafiz <ayaz.hafiz.1@gmail.com>; Author <author@noreply.github.com>
Тема: Re: [ayazhafiz/hgb] Change parser test strategy (#28)


@lukasberglund approved this pull request.

Cool I think this is great. Really nice to have test coverage back. I don't really know how we push this since it depends on the other PR. Can you handle that?

PR #29, opened 2020-03-17
Reordered functions in parser and lexer so that source files begin wi…

lukasberglund:
…th low level functions and end with high level functions


ayazhafiz:
I actually think the way we had it is fine. It’s probably better to put higher level functions at the top since that’s what the file is for. Everything else should just a logical grouping, in my opinion. So maybe it makes sense to place the precedence function on top of the infix parsing function.

lukasberglund:
Alright!

PR #34, opened 2020-03-22
Humanize parser error messages

### LINKED ISSUE #32: Specify expected symbol kinds in error messages
It is not very helpful to show symbols in error messages as the symbol name; it is more useful to represent the symbol as the token that generates it.

ayazhafiz:
This commit humanizes hgb's parser error messages by emitting errors
that are more actionable to end an end user than the state recorded by
the parser. For example, previously for the expression `5 + * 5!`, hgb
would output the error

```
Error: expected Plus, Minus, LParen, Number, Name, or StrBound; got Times
```

With this commit, the error message is nicer:

```
Error: expected an expression; got "*" at 4:5
```

Closes #32


ayazhafiz:
@lukasberglund this is still WIP (have to run rn) but relevant changes are in src/Errors. Updates to specs will be coming soon...

src/Parser.hs
```diff
@@ -28,8 +28,9 @@ closingDelim s =
   case s of
     Symbol.LParen -> Symbol.RParen
 
-wrapError :: ErrorType -> Either Error any
-wrapError err = Left (Error err)
+wrapExpectation :: Expectation -> Token -> Either Error any
+wrapExpectation expectation gotTok =
+  Left (Error (Expected expectation gotTok) (sourceSpan gotTok))
```
lukasberglund:
```suggestion
  Left $ Error (Expected expectation gotTok) (sourceSpan gotTok)
```

src/Token.hs
```diff
@@ -1,14 +1,12 @@
...
 
 data Token =
   Token
     { symbol :: Symbol
     , content :: String
-    -- Refers to the index of the first character of the token in the source code
-    , start :: Int
-    -- Refers to the index of the character after the last
-    , end :: Int
+    , sourceSpan :: Span
```
lukasberglund:
```suggestion
    , sourceSpan :: Span -- refers to the start and end of the token in the source code
```

src/Symbol.hs
```diff
@@ -50,3 +51,6 @@ data Symbol
   -- TODO(ayazhafiz): it would be nicer if we showed the expected symbol token rather than the
   -- symbol name.
   deriving (Eq, Show)
+
+symbolToStr :: Symbol -> String
+symbolToStr RParen = ")"
```
lukasberglund: Why not implement show?

ayazhafiz: This is actually a placeholder for now. I don’t want to implement show because not every symbol actually has a corresponding source code token (for example name doesn’t). I think what we should do is make a map of string -> symbol, use that in the Lexer, and make an inverted map to get text from a symbol as done here. I will do this in a follow up PR if that strategy sounds reasonable to you.

lukasberglund: Yup, sounds reasonable.

src/Parser.hs
ayazhafiz: Done

src/Token.hs
ayazhafiz: Done

ayazhafiz:
@lukasberglund this is ready; PTAL

src/Lexer.hs
```diff
@@ -9,10 +9,11 @@ import Prelude hiding (lex)
 import Symbol (Symbol)
 import qualified Symbol (Symbol(..))
 import Token (Token(..))
+import Types (Span(..), start, end)
 import Utils
 
 endOfFileToken :: Int -> Token
-endOfFileToken i = Token Symbol.EndOfFile "" i i
+endOfFileToken i = Token Symbol.EndOfFile "<EOF>" (Span i i)
```
lukasberglund: I don't really care what is in that string, but I was just wondering why we are putting something in it?

src/Types.hs
```diff
@@ -0,0 +1,13 @@
...
+
+data Span =
+  Span
+    { start :: Int -- index of the first character of a source code span
+    , end :: Int -- index of the character after the last of a source code span
+    }
+  deriving (Eq)
+
+instance Show Span where
+  show (Span start end) = show start ++ ":" ++ show end
```
lukasberglund: I'm worried that the name Types.hs is a bit confusing as it could be understood to refer to the types of the gingerbread programming languages (`int`, `double`, etc)

src/Error.hs
```diff
@@ -20,20 +22,30 @@ displayOptions all = concat all -- Option, AnotherOption, ..., or YetAnotherOpti
...
+data Expectation
+  = Expression
+  | ExpressionTerminator
+  | ExpressionOrDelimiter
+  | EnumerationListDelimiter [Symbol]
+  | CloseToken Symbol
+
+instance Show Expectation where
+  show Expression = "an expression"
+  show ExpressionTerminator = show (symbolToStr Symbol.ExprEnd)
```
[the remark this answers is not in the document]
ayazhafiz: Yes. To show the quotes.

src/Lexer.hs
ayazhafiz: In an error message we show the token contents. It would be nice to show some content

src/Types.hs
ayazhafiz: Would you prefer I move this to `utils` or a new module like `common`?

lukasberglund: Yeah I think utils would work.

ayazhafiz: Done

ayazhafiz:
ptal @lukasberglund

PR #35, opened 2020-03-22
Read programm text from stdin

ema-fox:
Makes it easier to interactively test out the lexer.

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -4,4 +4,5 @@ import qualified Lib as LibHgb
 
 main :: IO ()
 main = do
-    print (LibHgb.lex "=/=")
+    input <- getContents
+    print (LibHgb.lex input)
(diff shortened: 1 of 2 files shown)


app/Main.hs
```diff
@@ -4,4 +4,5 @@ import qualified Lib as LibHgb
 
 main :: IO ()
 main = do
-    print (LibHgb.lex "=/=")
+    input <- getContents
+    print (LibHgb.lex input)
```
lukasberglund:
```suggestion
    print $ LibHgb.parse $ LibHgb.lex input
```

ayazhafiz: Let’s just take a command line flag for parsing vs lexing. Nice to github “meet” you, @ema-fox.

lukasberglund:
Why can't this merge?

PR #36, opened 2020-03-22
Parse function decls, variable assign/decls

lukasberglund:
Now we can parse function declarations, variable assignments, and variable declarations.


src/Grammar.hs
```diff
@@ -1,9 +1,28 @@
...
+  , readType
+  ) where
 
 data Expression
   = Number Int
   | String String
-  | Variable String
+  | VariableRef String
+  | VariableDeclare
+      { variableName :: String
```
ayazhafiz:
```suggestion
      { name :: String
```

src/Grammar.hs
```diff
@@ -1,9 +1,28 @@
...
+  , Type(..)
+  , readType
+  ) where
 
 data Expression
   = Number Int
   | String String
-  | Variable String
+  | VariableRef String
+  | VariableDeclare
```
ayazhafiz:
```suggestion
  | VariableDeclaration
```

src/Grammar.hs
```diff
@@ -1,9 +1,28 @@
...
+      { variableName :: String
+      , variableType :: Type
```
ayazhafiz:
```suggestion
      , type :: Type
```

If this is a keyword, prefer `ty`.

src/Grammar.hs
```diff
@@ -1,9 +1,28 @@
...
+      , declaredVal :: Maybe Expression
```
ayazhafiz:
```suggestion
      , declaration :: Maybe Expression
```

src/Grammar.hs
```diff
@@ -1,9 +1,28 @@
...
+      }
+  | VariableAssign
```
ayazhafiz:
```suggestion
  | VariableAssignment
```

Could we just name this `Assignment`?

src/Grammar.hs
```diff
@@ -1,9 +1,28 @@
...
+      { variableName :: String
+      , assignedVal :: Expression
```
ayazhafiz:
```suggestion
      , assignment :: Expression
```

src/Grammar.hs
```diff
@@ -1,9 +1,28 @@
...
+      }
+  | FunctionDef
+      { definedFunction :: String
```
ayazhafiz:
```suggestion
      { name :: String
```

src/Grammar.hs
```diff
@@ -1,9 +1,28 @@
...
+      , params :: [Expression]
```
ayazhafiz: This should be `VariableRef`s, no?

src/Grammar.hs
```diff
@@ -1,9 +1,28 @@
...
+      , functionType :: Type
```
ayazhafiz:
```suggestion
      , type :: Type
```

or `ty`

src/Grammar.hs
```diff
@@ -1,9 +1,28 @@
...
+      , exprs :: [Expression]
```
ayazhafiz:
```suggestion
      , definition :: [Expression]
```

src/Grammar.hs
```diff
@@ -13,14 +32,58 @@ data Expression
       --   |    "+1"     |    (+ 1)    |
       --   | "foo(1, 2)" |  (foo 1 2)  |
       --   |-------------|-------------|
-      { name :: String
+      { calledFunction :: String
```
ayazhafiz: Why not just keep `name`?

src/Grammar.hs
```diff
@@ -13,14 +32,58 @@ data Expression
...
       , arguments :: [Expression]
       }
+  | Return Expression
+  deriving (Eq)
+
+data Type
+  = Int
+      { numBits :: Maybe Int
+      , signed :: Bool
+      }
```
ayazhafiz: Better to represent `int`, `int8`, ..., `uint`, `uint8`, ... separately. Or let's make `numBits` an enumeration.

src/Grammar.hs
```diff
@@ -13,14 +32,58 @@ data Expression
...
+
+readType :: String -> Type
+readType "void" = Grammar.Void
+readType "double" = Grammar.Double
+readType "int" = Grammar.Int Nothing True
+readType "uint" = Grammar.Int Nothing False
+readType ('i':'n':'t':numStr) = Grammar.Int bits True
+  where bits = Just (read numStr :: Int)  
+readType ('u':'i':'n':'t':numStr) = Grammar.Int bits False
+  where bits = Just (read numStr :: Int)
```
ayazhafiz: Why not implement `read` for `Type`?

src/Lexer.hs
```diff
@@ -128,7 +128,7 @@ lexOperatorOrPunctuation i str = (Token symbol tok i end, remainder)
         "*" -> (Symbol.Times, tok, rest)
         "/" -> (Symbol.Div, tok, rest)
         "%" -> (Symbol.Mod, tok, rest)
-        "!" -> (Symbol.ExprEnd, tok, rest)
+        "!" -> (Symbol.LineDelim, tok, rest)
```
ayazhafiz: :+1: thanks

src/Parser.hs
```diff
@@ -31,51 +30,144 @@ closingDelim s =
 wrapError :: ErrorType -> Either Error any
 wrapError err = Left (Error err)
 
+allowIfMatches :: [Symbol] -> [Token] -> Either Error [Token]
```
ayazhafiz:
```suggestion
assertMatches :: [Symbol] -> [Token] -> Either Error [Token]
```

src/Parser.hs
```diff
@@ -31,51 +30,144 @@ closingDelim s =
...
-  if | sym == Symbol.ExprEnd -> (expression :) <$> parse rest
-     | otherwise ->
-       wrapError ExpectedSymbol {expected = [Symbol.ExprEnd], actual = sym}
+parse tokens = fst <$> parseLines Symbol.EndOfFile tokens
 
-parseExpression :: Int -> [Token] -> Either Error (Expression, [Token])
-parseExpression precedence tokens = do
-  (left, afterLeft) <- parsePrefix tokens
-  parseInfix precedence left afterLeft
+parseLines :: Symbol -> [Token] -> Either Error ([Expression], [Token])
```
ayazhafiz:
```suggestion
parseLines :: Symbol -> Parser
```

src/Parser.hs
```diff
@@ -31,51 +30,144 @@ closingDelim s =
...
+parseLines = parseExprs [Symbol.LineDelim]
 
+parseExprs ::
+     [Symbol] -> Symbol -> [Token] -> Either Error ([Expression], [Token])
```
ayazhafiz:
```suggestion
     [Symbol] -> Symbol -> Parser
```

src/Parser.hs
```diff
@@ -31,51 +30,144 @@ closingDelim s =
...
-parse [Token Symbol.EndOfFile _ _ _] = Right []
-parse tokens = do
-  (expression, afterExpression) <- parseExpression 0 tokens -- parse first expression
-  -- Now, parse the rest
-  let (tok:rest) = afterExpression
-  let sym = symbol tok
-  if | sym == Symbol.ExprEnd -> (expression :) <$> parse rest
-     | otherwise ->
-       wrapError ExpectedSymbol {expected = [Symbol.ExprEnd], actual = sym}
+parse tokens = fst <$> parseLines Symbol.EndOfFile tokens
```
ayazhafiz: I'm confused by what this does. Could you explain and add a comment?

src/Parser.hs
```diff
@@ -31,51 +30,144 @@ closingDelim s =
...
 
-parseExpression :: Int -> [Token] -> Either Error (Expression, [Token])
-parseExpression precedence tokens = do
-  (left, afterLeft) <- parsePrefix tokens
-  parseInfix precedence left afterLeft
+parseLines :: Symbol -> [Token] -> Either Error ([Expression], [Token])
+parseLines = parseExprs [Symbol.LineDelim]
```
ayazhafiz: Also confused by what this does. I understand that the tokens from the call in `parse` are passed along, but what is the role of the `Symbol.EndOfFile` passed from `parse`?

src/Parser.hs
```diff
@@ -31,51 +30,144 @@ closingDelim s =
...
 --
 -- 2. Next, parse infix operators. These are operators that *do* rely on previous parsing contexts.
 --    Infix operators include the + and - functions.
--- 
+--
 -- [Top-Down Operator Precedence]: https://tdop.github.io/
 -- [Pratt Parsing]: https://journal.stuffwithstuff.com/2011/03/19/pratt-parsers-expression-parsing-made-easy/
+type Parser = ([Token] -> Either Error ([Expression], [Token]))
+
+type ParseFunction = ([Token] -> Either Error (Expression, [Token]))
```
ayazhafiz:
```suggestion
type ExpressionParser = ([Token] -> Either Error (Expression, [Token]))
```

Since we're parsing a single expression here, right?

src/Parser.hs
```diff
@@ -31,51 +30,144 @@ closingDelim s =
...
+parseEnum :: ([Symbol] -> ParseFunction) -> Symbol -> Symbol -> Parser
+-- parseEnum vagueParseFunction delim listOpen afterOpen
+-- Parses an enumerated list
+--   (data, "value")!
+--   ^                 | list open
+--                 ^   | list close
+--    ^^^^^^^^^^^^^^^  | afterOpen
+-- The difference between parseEnum and parseMultiple is that parseEnum
+-- does not require an enumerator at the end of the last expression whereas
+-- parseMultiple does. Do for example
```
ayazhafiz: Why not use the same function with a parameter to mark whether we should parse a last delimiter? It is not a big deal here, and I think it is probably better to keep those functions separate. Just something to think about though.

src/Parser.hs
```diff
@@ -31,51 +30,144 @@ closingDelim s =
...
+    parseMultiple' :: [Token] -> Either Error ([Expression], [Token])
+    parseMultiple' tokens
+      | (symbol $ head tokens) == terminator = Right ([], tokens)
+      | otherwise = do
+        (expr, afterExpr) <- parseFun tokens
+        let (delim:afterDelim) = afterExpr
+        (otherExprs, beyond) <- parseMultiple' afterDelim
+        return (expr : otherExprs, beyond)
+
+parseEnum :: ([Symbol] -> ParseFunction) -> Symbol -> Symbol -> Parser
```
ayazhafiz: Let's keep this named `parseEnumeration`. Also, why move it from its previous location?

src/Parser.hs
```diff
@@ -31,51 +30,144 @@ closingDelim s =
...
+parseMultiple :: ParseFunction -> Symbol -> Parser
+-- Parses multiple lines using a parsefunction until it hits a terminator.
+parseMultiple parseFun terminator = parseMultiple'
+  where
+    parseMultiple' :: [Token] -> Either Error ([Expression], [Token])
+    parseMultiple' tokens
+      | (symbol $ head tokens) == terminator = Right ([], tokens)
+      | otherwise = do
+        (expr, afterExpr) <- parseFun tokens
+        let (delim:afterDelim) = afterExpr
```
ayazhafiz: do we need to verify the delimiter is what we expect here?

src/Parser.hs
```diff
@@ -31,51 +30,144 @@ closingDelim s =
...
+  | otherwise = parseEnum' afterOpen
+  where
+    listClose = closingDelim listOpen
+    parseFunction = vagueParseFunction [delim, listClose]
+    parseEnum' :: [Token] -> Either Error ([Expression], [Token])
+    parseEnum' tokens = do
+      (item, afterItem) <- parseFunction tokens
+    -- Parse tokens after the enumerated item. There are two cases here:
+    --   1. the item is followed by a item delim
+    --   2. the item is followed by a list close
```
ayazhafiz: nit: indent

src/Parser.hs
```diff
@@ -31,51 +30,144 @@ closingDelim s =
...
-            [Symbol.LParen, Symbol.Number, Symbol.Name, Symbol.StrBound]
-        , actual = sym
-        }
+    case sym of
+      Symbol.LParen -> parseParen tokens
+      Symbol.Number -> parseNumber tokens
+      Symbol.Name -> parseName delims tokens
+      Symbol.StrBound -> parseString tokens
+      Symbol.Return -> parseReturn tokens
+      _ -> wrapError ExpectedSymbol {expected = possibleSymbols, actual = sym}
```
ayazhafiz: Why change this?

src/Parser.hs
ayazhafiz: I would rather not move it so that the diff of this PR doesn't increase. We can change the order in a different commit.

src/Grammar.hs
```diff
@@ -13,14 +32,58 @@ data Expression
...
+
 instance Show Expression where
   show (Number n) = show n
   show (String s) = show s
-  show (Variable v) = "$" ++ v
+  show (VariableRef v) = "$" ++ v
   show (Call name arguments) =
-    "(" ++ name ++ " " ++ unwords (map show arguments) ++ ")"
+    "(" ++ name ++ " " ++ displayList arguments ++ ")"
+  show (FunctionDef name params type' exprs) =
```
ayazhafiz:
```suggestion
  show (FunctionDef name params ty exprs) =
```

src/Grammar.hs
```diff
@@ -13,14 +32,58 @@ data Expression
...
+    "(fun " ++
+    name ++
+    " (" ++
+    displayList params ++
+    ") : " ++ show type' ++ " = " ++ displayList exprs ++ ")"
```
ayazhafiz:
```suggestion
    ") : " ++ show type' ++ " = (" ++ displayList exprs ++ "))"
```

ayazhafiz: So that we can get `(fun PRINT ((a : int)) : void = ($hey (+ 1 1)))`

src/Grammar.hs
lukasberglund: I got errors when I did this. It seems like haskell does not want overlapping fields for different constructors of the same type. That is, I can't have a field `name` for `VariableDef` if I also have the field `name` for `VariableRef`

src/Grammar.hs
lukasberglund: True

src/Grammar.hs
ayazhafiz: oh... that makes sense. Sounds good.

src/Parser.hs
lukasberglund: No, because now the `parseFunction` only stops at our specified delimiters.

src/Parser.hs
lukasberglund: I thought it was easier to read.

src/Grammar.hs
lukasberglund: I think `ty` is a bit nebulous as a name.

src/Parser.hs
ayazhafiz: I see. `parseFun` itself stops at a delimiter. Maybe add a comment here.

lukasberglund:
@ayazhafiz could you take a look at the coverage here? It's once again saying that fields in Error.hs and Grammar.hs aren't being accessed in the tests when they are.

src/Parser.hs
```diff
@@ -121,116 +178,142 @@ parsePrefix tokens@(tok:_)
...
+parseAssign :: [Delim] -> ParseFunction
+parseAssign delims (nameTok:assign:afterAssign) = do
+  (assignedVal, afterVal) <- parseExpr 0 delims afterAssign
+  -- goes with a precedence of zero, i think this is right but it could lead to some problems
+  return (Grammar.Assignment name assignedVal, afterVal)
+  where
+    name = content nameTok
+
+parseEnum :: ([Delim] -> ParseFunction) -> Delim -> Terminator -> Parser
+-- parseEnum vagueParseFunction delim listOpen afterOpen
```
ayazhafiz: let’s name this parseEnumeration as we will likely introduce an `enum` datatype.

src/Grammar.hs
```diff
@@ -1,9 +1,28 @@
...
 
 data Expression
   = Number Int
   | String String
-  | Variable String
+  | Reference String
+  | Declaration
+      { name :: String
+      , ty :: Type
+      , maybeAssignment :: Maybe Expression
```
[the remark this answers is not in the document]
lukasberglund: Because haskell dictates that if two fields have the same name they have to have the same type. Therefore since `assignment` is already an expression in the `Assignment` constructor I couldn't call it that.

test/Parser/ParserSpec.hs
```diff
@@ -106,21 +150,21 @@ spec = do
...
+        , "Error: expected an expression or \"!\", got \"<EOF>\" at ")
       , ( "invalid expression ending"
         , [ ("f(),", "3:4")
           , ("name,", "4:5")
           , ("4,", "1:2")
           , ("4 + 5,", "5:6")
           , ("(2 + 3),", "7:8")
           ]
-        , "Error: expected \"!\", got \",\" at ")
+        , "Error: expected an expression or \"!\", got \",\" at ")
```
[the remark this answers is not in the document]
lukasberglund: This would be tricky to code into the current parser. There would have to be some kind of check that the user has entered a terminator that is not the terminator we are currently using. We can do it, but it doesn't come naturally.

src/Grammar.hs
lukasberglund: What do you mean?

src/Grammar.hs
lukasberglund: Sorry I didn't see all these last time. Why make numBits an enumeration?

src/Grammar.hs
lukasberglund: I tried but kept running into errors for some reason....

src/Parser.hs
lukasberglund: Takes the first argument of the `([Expression], [Token])` Tuple returned by calling `parseLines` with `EndOfFile` as a terminator.

src/Parser.hs
lukasberglund: `Symbol.EndOfFile` designates where `parseLines/ parseExprs` should stop parsing.

src/Parser.hs
lukasberglund: True

src/Parser.hs
lukasberglund: One could do that, but I feel like you would end up writing the same amount of code, right?

src/Parser.hs
```diff
@@ -5,68 +5,125 @@ module Parser
...
+parseExpr :: Int -> [Delim] -> ParseFunction
+parseExpr precedence delims tokens =
+  if symbol (head tokens) == Symbol.Func
+    then parseFunctionDeclaration $ tail tokens
+    else do
+      (left, afterLeft) <- parsePrefix delims tokens
+      parseInfix precedence delims left afterLeft
+
+parseFunctionDeclaration :: ParseFunction
+parseFunctionDeclaration tokens =
```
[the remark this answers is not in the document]
lukasberglund: I switched it back to that!

src/Grammar.hs
ayazhafiz: Because the number of bits we can have is well defined. It is not the case that we can have `numBits 12`, for example.

lukasberglund:
So how do we merge this? Do we just merge the other one??

ayazhafiz:
Merge the other one into this, then this into master

Отправлено: Monday, March 30, 2020 3:05:21 AM
Копия: hafiz <ayaz.hafiz.1@gmail.com>; Mention <mention@noreply.github.com>
Тема: Re: [ayazhafiz/hgb] Parse function decls, variable assign/decls (#36)


So how do we merge this? Do we just merge the other one??

lukasberglund:
Okay I'm good merging. Are you?

PR #41, opened 2020-03-29
Started on Eval. Wrote evalInt.

lukasberglund:
Is this what you had in mind as eval?

### DIFF
--- a/src/Eval.hs
+++ b/src/Eval.hs
@@ -1,0 +1,14 @@
+module Eval where
+
+import Grammar
+
+evalInt :: Expression -> Int
+evalInt (Number n) = n
+evalInt (Call function args) =
+  case function of
+    "+" -> a + b
+    "-" -> a - b
+    "*" -> a * b
+    "/" -> a `div` b
+    "%" -> a `mod` b
+  where [a, b] = map evalInt args


src/Eval.hs
```diff
@@ -0,0 +1,17 @@
+module Eval where
+
+import Grammar
+
+evalInt :: Expression -> Int
```
ayazhafiz:
```suggestion
evalNumber :: Expression -> Int
```

lukasberglund: But what about floats?

ayazhafiz: i'd prefer to maintain consistency with our grammar names. we can add support for floats later on

lukasberglund: But evalInt has to return int since Haskell is strongly typed. Are you suggesting we change the name back when we add floats?

ayazhafiz: Let's see what happens with your change to the `Number` grammar first. If we move to a string this won't work anyway.

lukasberglund: So what do you think?

ayazhafiz: Since we chose to change the grammar of a number to hold a string, this won't work yet. We need to add an intermediate step to translate the number into a particular type before we can evaluate it. This will actually require some work... so let's do that first.

Basically given some expression like `1+1`, right now we can parse it to `Number(1) + Number(1)`. We need some step that can infer this expression to be `Int(1) + Int(1)`. In a later step we can add validation and errors for mismatched types as well; I think here it is enough to make inferences for the `+, -, /, *, %` functions. Then later on we can check that the parameters passed to the functions are as expected. Does that make sense?

So basically I suggest the flow right now should be

```
Parser -> Type Resolution --> Type Checking (add later) --> Eval
```

lukasberglund: Makes sense. If we are doing lazy evaluation by default we only have to evaluate these expressions when it's necessary. Does this factor into what we should be doing now?

Also, what would type resolution look like?

I'm imagining some process where we add a type to each expression-branch in the AST. Is that right?

ayazhafiz: Not sure, I'm actually not super familiar with how to implement lazy evaluation on the compiler side. My assumption is that it will only be for the evaluation phase, not anything before (type reso, checking, etc)

For type resolution, yes. It can just be a granular grammar with type fields.

lukasberglund:
I ~~added a naive implementation of a typechecker~~ singlehandedly solved typechecking. PTAL!

lukasberglund:
I just wanted to know if you thought my approach is good.

src/Eval.hs
```diff
@@ -0,0 +1,44 @@
...
+
+driveEval :: String -> Either Error String
+driveEval s = pure (Lexer.lex s) >>= parse >>= (showResult . head)
+
+showResult :: Expression -> Either Error String
+showResult expr = do
+  ty <- expressionType expr
+  case ty of
+    Int _ _ -> return $ show $ evalInt expr
+    Double -> return $ show $ evalDouble expr
```
ayazhafiz, line 17: should probably go in the test file?

ayazhafiz: I guess `showResult` could be called `eval` where we do the type checking then route to the appropriate eval fn

src/TypeCheck.hs
```diff
@@ -0,0 +1,42 @@
...
+expressionType expr = fmap head (expressionTypes expr)
+  where
+    expressionTypes :: Expression -> Either Error [Type]
+    expressionTypes (Number s) =
+      if '.' `elem` s
+        then Right [Double]
+        else Right numTypes
+    expressionTypes (String _) = Right [Str]
+    expressionTypes (Char _) = Right [CharTy]
+    expressionTypes (Call name [a, b]) =
```
ayazhafiz, line 28: can you add a todo to support arbitrary # parameters

src/TypeCheck.hs
```diff
@@ -0,0 +1,42 @@
...
+      pure operationTypes >>= filterTypes a >>= filterTypes b
+      where
+        operationTypes
+          | name `elem` ["+", "-", "*", "/"] = numTypes
+          | name == "%" = allInts
```
ayazhafiz, line 33: Actually maybe we should just do the typechecking for numbers and calls rn since that's what eval cares about

src/Eval.hs
lukasberglund: So do we want eval to always return a string?

src/TypeCheck.hs
lukasberglund: You mean before we do eval?

src/Eval.hs
ayazhafiz: No I think the return type is fine I just think the name could be better.

src/TypeCheck.hs
ayazhafiz: yes

hgb.cabal
```diff
@@ -1,81 +0,0 @@
-cabal-version: 1.12
```
ayazhafiz, line 1: ?? why delete this?

src/Eval.hs
```diff
@@ -0,0 +1,51 @@
...
+evalDouble :: Expression -> Float
+evalDouble (Number n) = read n :: Float
+evalDouble (Call "+" [x]) = evalDouble x
+evalDouble (Call "-" [x]) = -evalDouble x
+evalDouble (Call function args) =
+  case function of
+    "+" -> a + b
+    "-" -> a - b
+    "*" -> a * b
+    "/" -> a / b
```
ayazhafiz, line 49: What about mod for doubles?

ayazhafiz: Ah it's only ints

src/TypeCheck.hs
```diff
@@ -0,0 +1,46 @@
...
+      case types `intersect` exprTypes of
+        [] -> wrapTypeError types expr
+        ts -> return ts
+
+functionTypes :: String -> Either Error [Type]
+functionTypes "+" = Right numTypes
+functionTypes "-" = Right numTypes
+functionTypes "*" = Right numTypes
+functionTypes "/" = Right numTypes
+functionTypes "%" = Right allInts
```
ayazhafiz, line 46: What if we make a registry of function names to their type signature and implementation? It will make it easier to typecheck custom functions later and have a common place for the type signature and implementation rather than splitting between typecheck and eval. Just something to think about, we can do it in another PR

src/TypeCheck.hs
```diff
@@ -0,0 +1,46 @@
...
+    expressionTypes :: Expression -> Either Error [Type]
+    expressionTypes (Number s) =
+      if '.' `elem` s
+        then Right [Double]
+        else Right numTypes
+    expressionTypes (String _) = Right [Str]
+    expressionTypes (Char _) = Right [CharTy]
+    expressionTypes (Call name args) = do
+      funTypes <- functionTypes name
+      foldrM filterTypes funTypes args
```
ayazhafiz, line 31: can you add a TODO that this should (1) return the return type of the function and (2) should compare each argument with each parameter type rather than the list of all parameter types of the function

src/TypeCheck.hs
```diff
@@ -0,0 +1,46 @@
...
+
+import Data.Foldable (foldrM)
+import Data.List (intersect)
+import Error
+import Grammar (Bits(..), Expression(..), Type(..))
+import Utils
+
+wrapTypeError :: [Type] -> Expression -> Either Error a
+wrapTypeError expectedTypes expression =
+  Left $ Error (TypeError expectedTypes expression) (Span 0 0)
```
ayazhafiz, line 13: TODO: fix the span?

lukasberglund: I can add that as a TODO! I can't actually fix the span yet since expressions aren't linked to code.

hgb.cabal
lukasberglund: Ill add it back

src/TypeCheck.hs
lukasberglund: That could work.

src/TypeCheck.hs
lukasberglund: Ah true, lol I actually thought that this was was the function was gonna do. I honestly don't know what I was thinking.

ayazhafiz:
@lukasberglund do you want a review on this?

PR #42, opened 2020-03-30
 Minor refactoring of and coverage bump for #36 

ayazhafiz:


src/Parser.hs
```diff
@@ -5,15 +5,16 @@ module Parser
...
 import Error (Error(..), ErrorType(..), Expectation(..))
-import Grammar (Expression, Type, readType)
+import Grammar (Expression, Type)
 import qualified Grammar
 import Symbol (Symbol(..))
 import qualified Symbol
 import Token (Token(..))
 import Utils
 
+newtype Void = Void Void
```
lukasberglund: Can you explain what's going on here? What do the two Voids refer to?

src/Parser.hs
```diff
@@ -83,35 +82,33 @@ parseExpr precedence delims tokens =
 
 parseFunctionDeclaration :: ExpressionParser
 parseFunctionDeclaration tokens = do
-  expectedStart `assertMatches` tokens
+  tokens <- expectedStart `assertMatches` tokens
```
lukasberglund: What's the point in renaming tokens here?

src/Parser.hs
```diff
@@ -83,35 +82,33 @@ parseExpr precedence delims tokens =
...
   let nameTok:lParen:paramsAndBeyond = tokens
   (params, afterParams) <- parseDeclarations paramsAndBeyond
-  expectedMid `assertMatches` afterParams
+  tokens <- expectedMid `assertMatches` afterParams
```
lukasberglund: And here

src/Parser.hs
ayazhafiz: sorry, i should remove this

src/Parser.hs
ayazhafiz: just to satisfy coverage, because `assertMatches` returns tokens on the right

PR #45, opened 2020-03-31
Change type in Grammar.Number constructor

lukasberglund:
Change type in Grammar.Number constructor to be agnostic between int and float. Why? So that we only read the type of numbers once we know whether they should be interpreted as `float` or `int` or `uint` etc.


src/Grammar.hs
```diff
@@ -56,7 +56,7 @@ instance Read Type where
       readType ('u':'i':'n':'t':maybeBits) = Int (readMaybe maybeBits :: Maybe Bits) False
 
 data Expression
-  = Number Int
+  = Number String
```
ayazhafiz: Could we differentiate between float and int at this step?

lukasberglund: This is a tough question, if we do, how would we deal with something like this?
```
a : double = 1!
a + 5!
```
If we had already parsed the 5 as int then we'd have to convert it afterwards. Alternatively we could evaluate is as it comes.

ayazhafiz: Yeah. We might not have enough information at parse time to decide what the type of the number is. I think there are a few alternatives:

- Store as a string and infer the type later on, before we call the evaluator/compiler
- Require numbers to be strongly typed. So here, we would assume the 5 is an int, and the only way this would be valid is if you did something like `a + 5f32` or `a + 5.`
- Add a "hint" for the expected number type here. Then we can do an implicit type conversion if the types are compatible without having to have an inference step before the evaluator. But this is a short-term solution, we will need a type binder later on anyway

lukasberglund: Do some language actually let you / make you write something like `a + 5f32`? 

To me that seems a bit excessive. We could make numbers be typed in the sense that anything with a dot is always some kind of float and everything without a dot is always some kind of int, but I personally get annoyed when I have to write something like `5.` cause it always seems hacky. 

Also I don't understand the third option, where would the hint be added?

ayazhafiz: I agree with you. The work should be on hgb to infer the type

The hint would be added in our grammar when we parse the number. Then we read it later on, either during evaluation of the "real" type or program eval

lukasberglund: I feel like we don't really need a hint because if we save the number as a raw string and we have put into context via our syntax tree then we aren't losing any info we have on the type of the number.

PR #46, opened 2020-03-31
Support escape characters

### LINKED ISSUE #43: Allow the use \" inside of strings

ayazhafiz:
Adds support for escaping

- string delimiter (")
- tabs
- newlines

Closes #43


src/Lexer.hs
```diff
@@ -33,19 +35,38 @@ lex' i all@(x:xs)
...
+-- There is nothing in the escape sequence, so just return the escape character.
+-- TODO: this should actually be a lexing error.
+lexEscapeSequence [] = (backslash, "")
+lexEscapeSequence (c:rest) = (content, rest)
+  where
+    content =
+      case c of
+        'n' -> '\n'
+        't' -> '\t'
+        other -> other
```
ayazhafiz, line 69: I want to reuse this for lexing `char`s in a next PR

src/Lexer.hs
```diff
@@ -33,19 +35,38 @@ lex' i all@(x:xs)
...
+    strCloseToken =
       Token Symbol.StrBound [delim] (Span contentEnd (contentEnd + 1))
+    lexStr' :: Int -> String -> String -> (Int, String, String)
+    lexStr' i content [] = (i, content, [])
+    lexStr' i content xs@(c:xs')
+      | c == delim = (i, content, xs)
+      | c == backslash =
+        let (char, rest) = lexEscapeSequence xs'
+         in lexStr' (i + 1) (content ++ [char]) rest
+      | otherwise = lexStr' (i + 1) (content ++ [c]) xs'
```
lukasberglund, line 57: Can we remove the first string argument and just return  something like `(end, char : moreContent, afterContent)` where `(end, moreContent, afterContent) = lexStr' (i + 1) rest`? 

This would also improve performance since we would no longer be adding a char to the end of content via the `++` operator which is an O(n) operation.

src/Lexer.hs
lukasberglund, line 69: When I type `echo "he\llo"` in bash it outputs `he\llo` which is the same as when I type `echo "he\\llo"` in bash. So it seems like bash reads a `\` followed by a non-escape character just like `\\` followed by the same character. Python does the same thing. This makes me think we should do that too. 

In this context, I think we should return escape characters like `'` and `"` as is but for non-escape characters we should return the backslash along with the character, just like bash and python. So for example we should return `\g` as `\g` just as it came in. What do you think?

src/Lexer.hs
ayazhafiz: i like this, thanks

src/Lexer.hs
ayazhafiz: Well, on the other hand node does elide the `\` in an escaped sequence:

```
❯ node
Welcome to Node.js v13.8.0.
Type ".help" for more information.
> 'hel\lo'
'hello'
```

And haskell reports a lexical error:

```
*Main Error Grammar Lexer Lib Parser Symbol Token Utils> "he\llo"

<interactive>:3:5: error:
    lexical error in string/character literal at character 'l'
```

So it seems like the behavior is not standardized. Personally I am more inclined to "escape" the character and remove the `\`, because in a general context that is what the escape character should be used for, but I am not strongly attached to this idea. let me know what you prefer

lukasberglund: I guess my N=2 experiment was not statistically significant. Fair enough, I don't have a strong attachment either.

Issue #38, opened 2020-03-23
Add char, array, and string (char array) type to gingerbread.


lukasberglund:
As of right now the lexer does not lex keywords like "string" or "char". And the parser does not deal with chars, arrays, or strings.

ayazhafiz:
Who is this assigned to? You or I?

lukasberglund:
I can do it. You wanna do it?

ayazhafiz:
Do char or array and I will do the other. Let's talk about strings. Are we sure that we want to represent strings as char arrays?

ayazhafiz:
Talked offline, we will proceed with char arrays

lukasberglund:
Actually can we reopen this char array thing? I was just thinking about how inconvenient arrays are in c/ c++, especially if you want to extend them. Do you think if we make strings be char arrays we risk making them hard to extend?

ayazhafiz:
well we can provide our own wrapper of a string class so that we do not actually expose strings as chars arrays, that’s just the representation inside. what do you think?

________________________________
От: lukasberglund <notifications@github.com>
Отправлено: Wednesday, April 1, 2020 9:56:08 AM
Кому: ayazhafiz/hgb <hgb@noreply.github.com>
Копия: hafiz <ayaz.hafiz.1@gmail.com>; Comment <comment@noreply.github.com>
Тема: Re: [ayazhafiz/hgb] Add char, array, and string (char array) type to gingerbread. (#38)


Actually can we reopen this char array thing? I was just thinking about how inconvenient arrays are in c/ c++, especially if you want to extend them. Do you think if we make strings be char arrays we risk making them hard to extend?

lukasberglund:
Yeah sounds good.

lukasberglund:
In the Gingerbread doc we had List syntax be something like this
```
arr : [Int; 4] = [3, 2, 5, 1]
```
I think the `[Int; 4] ` isn't very nice and could use some improvement. What do you think?

ayazhafiz:
what do you think are some ways we could improve it?

Maybe
```
Int[4]
Int<4>[]   # useful if we add support for generics later on
[Int]<4>
```

lukasberglund:
Those seem like good suggestions. How is the second one useful generics?

lukasberglund:
I kinda like the third option because it could allow people to create arrays that don't necessarily have a max size. What do you think of allowing that in gingerbread?

ayazhafiz:
[...]

PR #49, opened 2020-04-06
Add support for lexing chars

ayazhafiz:
This commits adds the `Char` and `CharBound` symbols, and support for
tokenizing chars. Chars that do not contain exactly one character are
considered invalid. Escaped characters are also handled.

This commit also backfills a fix for proper spans of escaped characters,
which can span more than one character (as in the case of "\t") but were
previously recorded as only spanning one.

Part of #38


src/Lexer.hs
```diff
@@ -37,33 +38,54 @@ lex' i all@(x:xs)
...
+  | ch == backslash =
+    let (escChar, len, rest) = lexEscapeSequence xs'
+        (end, moreContent, afterContent) = lexStrToDelim (i + len) rest delim
+     in (end, escChar : moreContent, afterContent)
+  | otherwise =
+    let (end, moreContent, afterContent) = lexStrToDelim (i + 1) xs' delim
+     in (end, ch : moreContent, afterContent)
+
+lexEscapeSequence :: String -> (Char, Int, String)
 -- There is nothing in the escape sequence, so just return the escape character.
```
lukasberglund, line 85:
```suggestion
-- Lexes an escape sequence returning the escaped character, the length of the escape sequence and the remaining code.
-- There is nothing in the escape sequence, so just return the escape character.
```

PR #53, opened 2020-04-06
Parse chars

ayazhafiz:
Part of #38


src/Parser.hs
```diff
@@ -295,8 +301,21 @@ parsePrefixCall delims (tok:rest) = do
   return (Grammar.Call (content tok) [expr], afterExpr)
 
 parseString :: ExpressionParser
-parseString (_:contentToken:_:afterString) =
-  Right (Grammar.String (content contentToken), afterString)
+parseString (opener:contentToken:closer:afterString) =
+  case symbol closer of
+    Symbol.StrBound ->
+      Right (Grammar.String (content contentToken), afterString)
+    _ -> wrapError (UnterminatedLiteral Symbol.String) (opener, contentToken)
```
lukasberglund, line 308: Do you think once we implement errors in the lexer we should add this to lex errors? We can keep it for now, just wondering.

ayazhafiz: Yes

PR #55, opened 2020-04-07
Add parser support for arrays 

lukasberglund:
#38


test/Parser/ParserSpec.hs
```diff
@@ -40,6 +40,7 @@ spec = do
         , "foo:uint64 = 10 + 20 * 30!"
         , "(declare ($foo : uint64) (+ 10 (* 20 30)))")
       , ("with arbitrary whitespace", "a  : \t void!", "(declare ($a : void))")
+      , ("with an array type", "a : [int]!", "(declare ($a : [int]))")
```
ayazhafiz, line 43: can you add an assignment test too?

src/Grammar.hs
```diff
@@ -60,6 +62,7 @@ instance Read Type where
 data Expression
   = Number String
   | String String
+  | List [Expression]
```
ayazhafiz, line 65: I don't think an expression should be able to be a list of expressions.

test/Parser/ParserSpec.hs
```diff
@@ -36,10 +42,14 @@ spec = do
       , ( "with an initializer"
         , "foo:uint64 = 10!"
         , "(declare ($foo : uint64) 10)")
+      , ( "with an initializer and an array type"
+        , "a : [double] = []!"
+        , "(declare ($a : [double]) [])")
```
ayazhafiz, line 47: nit: put this under `describe "arrays"`

src/Grammar.hs
lukasberglund: This is to initialize arrays.

PR #57, opened 2020-04-09
Parse string type

ayazhafiz:
Part of #38


ayazhafiz:
We have built a good parser :) changes like this should be trivial to add.

Issue #52, opened 2020-04-06
Add parser support for if statements.


lukasberglund:
We need to handle if statements! Here is the syntax we had in mind.

```
if str.empty {
        0!
} elseif str[0] == chr {
       1 + COUNT(str.drop_first, chr)!
} else {
        COUNT(str.drop_first, chr)!
} }!
```

PR #59, opened 2020-04-19
Support boolean and boolean operators

ayazhafiz:
Adds the `bool` type, `true` and `false` values, and parser support for
the prefix boolean operator `not` and infix operators `and`, `or`, `in`,
`==`, `=/=`, `<`, `<=`, `>`, `>=`.

Prepares for parsing conditional branches in #52.


src/Parser.hs
```diff
@@ -189,7 +214,7 @@ parseInfix precedence delims = parseInfix'
   where
     parseInfix' left tokens@(operatorTok:afterOperator)
       | operatorSym `elem` delims = Right (left, tokens)
-      | operatorSym `elem` infixOperators =
+      | newPrecidence > 0 =
```
lukasberglund, line 217: Ooh I see why you expanded `infixOperatorPrecidence`. I am not too sure about this because this doesn't make it clear that `newPrecidence > 0` is actually asking whether it is an infix operator. I think the better option would be to make  `infixOperatorPrecidence` return `Maybe Int` and then this conditional would be `newPrecidence != Nothing` or something of the sort.

ayazhafiz: Fixed :)

ayazhafiz:
@lukasberglund ptal

PR #61, opened 2020-04-20
Parse conditional expressions

### LINKED ISSUE #52: Add parser support for if statements.

ayazhafiz:
Adds support for parsing `if` expressions, with optional `elseif` and
`else` blocks.

Closes #52


ayazhafiz:
Waiting on #59.

src/Parser.hs
```diff
@@ -15,26 +17,74 @@ import Utils
...
+  where
+    returnConflict ::
+         [Symbol]
+      -> [Token]
+      -> [Token]
+      -> Either (Symbol, Token) ([Token], [Token])
+    returnConflict [] matched rest = Right (matched, rest)
+    returnConflict (x:xs) matched (y:ys)
+      | x == symbol y = returnConflict xs (matched ++ [y]) ys
+      | otherwise = Left (x, y)
```
lukasberglund, line 82:
```suggestion
    returnConflict ::
         [Symbol]
      -> [Token]
      -> Either (Symbol, Token) ([Token], [Token])
    returnConflict [] tokens = Right ([], rest)
    returnConflict (x:xs) (y:ys)
      | x == symbol y = first (x:) <$> returnConflict xs ys
      | otherwise = Left (x, y)
```

src/Parser.hs
```diff
@@ -15,26 +17,74 @@ import Utils
...
   Left $
   Error errorTy (Span (start $ sourceSpan startTok) (end $ sourceSpan endTok))
 
+assertMatches :: [Symbol] -> [Token] -> Either Error ([Token], [Token])
+-- Checks whether a list of matches a list of symbols at its start.
+-- If not, returns an error with for the token that did not match.
+-- Otherwise, returns a tuple of matched tokens and tokens after the match.
+assertMatches symbols tokens =
+  returnConflict symbols [] tokens &
+  first (\(expected, actual) -> makeExpectation [Symbol expected] actual)
```
lukasberglund, line 72: This is def neat, but do you feel like we are making our code too hard to understand in an effort to be as concise as possible?

ayazhafiz: Maybe, but I don't think this is too complicated. The only part that I think may be confusing is that we match against a list of expected symbols. What we could do instead is have a function that asserts the first token of a list is of an expected symbol. But I think without this we would want it anyway, since branching in the parsing functions could get nasty fast

PR #62, opened 2020-04-24
Simplified Conditionals

lukasberglund:
This stuff should have been in my code review but I didn't think about it until yesterday. Goes to show I should put more thought into code reviews. I'll be better about it next time.


ayazhafiz:
> Goes to show I should put more thought into code reviews. I'll be better about it next time.

No worries, this is an iterative process and the point is to be able to fixup code that we write earlier :)

src/Parser.hs
```diff
@@ -126,35 +126,28 @@ parseExpr precedence delims tokens =
...
+  let terminator:afterTerminator = afterIf
   (else', afterConditional) <-
-    case symbol $ head elseTokens of
-      Symbol.Else -> do
-        (exprs, afterExpressions) <- parseExpressionList (tail elseTokens)
-        Right (Just exprs, afterExpressions)
-      _ -> Right (Nothing, afterElseIf)
-  return (Grammar.Conditional if' then' elseIfs else', afterConditional)
+    case symbol $ head afterTerminator of
+      Symbol.ElseIf -> do
```
ayazhafiz, line 133: I am thinking we should change the syntax of `elseif` to be `else if`, which would make the way we express the grammar fall in line with the syntax of the language.

lukasberglund: True. Let's do it. So `else if` would be syntactic sugar for `else { if`

lukasberglund: Let's add it as an issue

ayazhafiz: Yep

PR #65, opened 2020-04-28
Changed names in Grammar and added number types

lukasberglund:
Changed naming of Expressions and Types in Grammar; created seperate types for Int and Double in Grammar. 

The motivation behind creating Double and Float types is that in the type theory book, evaluators worked by converting more complex expressions into simpler expressions step by step until they became values. That means we have to have values be expressions, so a generic number type is not enough.


src/Grammar.hs
```diff
@@ -30,67 +30,68 @@ instance Read Bits where
...
-  | Str
-  | Bool
-  | Void
-  | Array Type
+  | TDouble
+  | TChar
+  | TStr
+  | TBool
+  | TVoid
+  | TArray Type
```
ayazhafiz, line 42: :+1:

src/Grammar.hs
```diff
@@ -30,67 +30,68 @@ instance Read Bits where
...
-  | Char Char
-  | Reference String
-  | Declaration
+  = EInt Int
+  | EDouble Float
+  | EString String
+  | EArray [Expression]
+  | EChar Char
+  | EReference String
+  | EDeclaration
```
ayazhafiz, line 78: I would prefer something like `IntExp`. What do you think?

lukasberglund: How about `IntExpr` since `Exp` could be `Expectation`?

ayazhafiz:
sounds good

lukasberglund:
@ayazhafiz what do you think of the naming of types by adding the `T` prefix?

ayazhafiz:
I like it.

ayazhafiz:
Rebasing and merging 🚀

PR #66, opened 2020-04-28
Distinguish statements and expressions

### LINKED ISSUE #40: Parser does not distinguish between declarations and expressions
In terms of grammar, declarations and expressions are not interchangeable. For instance, you shouldn't be able to add a variable declaration to a number like so `3 + a : Int`. 

I would recommend creating a type `Statement` and declaring a bit like this:
```haskell
data Statement = Expression Expression | Declaration Declaration
```
where `Declaration` and `Expression` are each their own datatype. Then we refactor the parser around that.

ayazhafiz:
Gingerbread is an expression-based language whose control flow is guided
by statements. Currently, we parse all Gingerbread programs as a series
of expressions (which are evaluatable to a result), but in fact we
should parse programs as a series of statements (containers for
expression evaluations) to represent this control flow.

This commit distinguishes `Expression` and `Statement` grammars, and
moves the parser module to parse programs as a list of `Statement`s
rather than `Expression`s. In this commit, statements include expression
statements, variable declarations, and function declarations.

Because a sequence of statements in a block can serve as an expression
itself, this commit introduces a new expression: `BlockExpr`. A
`BlockExpr` consists of a series of statements followed by an
optional expression; the value of the `BlockExpr` is the value of
the final expression, or the unit type if there is no final expression.
`BlockExpr`s are useful as function or conditional bodies.
Following is an example listing of the values of a `BlockExpr`:

```gb
a : int  = if true { 1  } else { 2  } => a = 1
a : void = if true { 1; } else { 2; } => a is void
```

This commit also adds some TODOs to fix up inconsistencies in the
grammar introduced by the statement/expression differentiation. Most
notably, final expressions in `BlockExpr`s are currently not parsed.

Closes #40.


src/Grammar.hs
```diff
@@ -100,16 +93,30 @@ data Expression
...
+  | BlockExpr
+      -- Represents a brace-delimited block.
+      -- A block contains a series of statements and an optional final expression. If present, the
+      -- value of the final expression is returned from the block; otherwise, the return value of
+      -- the block is the unit type.
+      --   { a : int = 5! a + 1 }  => 6
+      --   { a : int = 5! a + 1! } => _
+      { statements :: [Statement]
+      , finalExpression :: Maybe Expression
+      }
```
lukasberglund, line 119: Cool

src/Grammar.hs
```diff
@@ -153,4 +153,33 @@ instance Show Expression where
             (if kind == "else"
                then ""
                else " then") ++
-            " (" ++ displayList exprs ++ ")"
+            " " ++ displayList exprs
+  show (BlockExpr stmts Nothing) = "{" ++ displayList stmts ++ "}"
+
+data Statement
+  = ExpressionStatement Expression
```
lukasberglund, line 160: When does it ever make sense to use an `ExpressionStatement`?

src/Parser.hs
```diff
@@ -70,17 +71,19 @@ assertMatches expectedSymbol (first:rest) =
     then return (first, rest)
     else wrapExpectation [Symbol expectedSymbol] first
 
-assertExpressionList :: [Expression] -> [Token] -> Either Error ()
-assertExpressionList [] rest = wrapExpectation [Expression] $ head rest
-assertExpressionList _ _ = Right ()
+assertNonNullList :: [a] -> Expectation -> [Token] -> Either Error ()
```
lukasberglund, line 74:
```suggestion
assertNonEmpty :: [a] -> Expectation -> [Token] -> Either Error ()
```
Maybe this is better (?)

src/Parser.hs
```diff
@@ -70,17 +71,19 @@ assertMatches expectedSymbol (first:rest) =
...
+assertNonNullList [] expectedItem rest = wrapExpectation [expectedItem] $ head rest
+assertNonNullList _ _ _ = Right ()
 
 type Delim = Symbol
 
 type Terminator = Symbol
 
-type ExpressionParser = ([Token] -> Either Error (Expression, [Token]))
+--                    input              parsed  remainder
+type UnitParser a = ([Token] -> Either Error (a, [Token]))
```
lukasberglund, line 83: Cool

src/Parser.hs
```diff
@@ -93,78 +96,103 @@ type MultiExpressionParser = ([Token] -> Either Error ([Expression], [Token]))
...
 -- Parses a program by parsing multiple lines of expressions until EOF is hit.
 parse tokens = fst <$> parseLines Symbol.EndOfFile tokens
 
-parseLines :: Terminator -> MultiExpressionParser
--- Parses multiple lines in program unitl some terminator is hit.
-parseLines = parseMultiple (parseExpr 0 [Symbol.LineDelim])
+parseLines :: Terminator -> ListParser Statement
+-- Parses lines in a program until some terminator is hit.
+-- A line is a statement terminated by a statement delimiter.
+parseLines = parseMultiple parseStmt
```
lukasberglund, line 107: I like what the redefinition of names does to the readability of the syntax here :)

src/Parser.hs
```diff
@@ -93,78 +96,103 @@ type MultiExpressionParser = ([Token] -> Either Error ([Expression], [Token]))
...
 
-parseMultiple :: ExpressionParser -> Terminator -> MultiExpressionParser
--- Parses multiple lines using an ExpressionParser some terminator is hit.
-parseMultiple expressionParser terminator = parseMultiple'
+parseMultiple :: UnitParser a -> Terminator -> ListParser a
```
lukasberglund, line 108: Ahhh

src/Parser.hs
```diff
@@ -93,78 +96,103 @@ type MultiExpressionParser = ([Token] -> Either Error ([Expression], [Token]))
...
+  return (Grammar.ExpressionStatement expr, afterExpr)
 
-parseExpr :: Int -> [Delim] -> ExpressionParser
+parseExpr :: Int -> [Delim] -> UnitParser Grammar.Expression
+-- Parses an expression.
 parseExpr precedence delims tokens =
   case symbol (head tokens) of
-    Symbol.Func -> parseFunctionDeclaration $ tail tokens
-    Symbol.If -> parseIfBlock $ tail tokens
+    Symbol.If -> parseConditional tokens
```
lukasberglund, line 147: If we make conditionals expressions, are we limiting the applicability of our language? For instance, how would one write something like

```
if is_even(num) {
    print("even!")!
} 
```

lukasberglund: Maybe to remedy this we could add an option that calls parseConditional in `parseStmt` and somehow make `parseConditional` able to return either an expression or a statement?

src/Grammar.hs
ayazhafiz: Most statements in Gingerbread will be expression statements. For example, `foo();`, `if true { do1(); }`, etc. are all expression statements.

src/Parser.hs
ayazhafiz: This is where `ExpressionStatements` become invaluable 🙂 

In most contexts, a conditional can be evaluated as an expression:

```
let a : int = if true { 1 } else { 2 };
```

In a non-received context (i.e. when the conditional result is not used as an input for another expression or statement), the conditional itself becomes a statement.

```
if true { print("true")! }  # now a statement
```

Should I add some test cases to better express this?

src/Grammar.hs
lukasberglund: Oh right, because our expressions are not pure.

src/Parser.hs
lukasberglund: Ahhh nice.

ayazhafiz:
Not merging until #67 is merged and I add tests to better show how conditionals can be expressions.

PR #67, opened 2020-04-28
Add fix options to scripts and squashed branch script

ayazhafiz:
`scripts/squashed` is added to ensure that branches consist of one
commit before they are rebased to master on a PR. Is it pedantic?
Definitely.


scripts/hindent
```diff
@@ -0,0 +1,23 @@
...
+#
+# Available options:
+#   --fix   Fix hindent errors
+set -e
+
+FILES=(src/*.hs test/**/*.hs)
+ACTION="${1:-}"
+case "$ACTION" in
+"")
+  hindent --validate \
```
lukasberglund, line 13: When is `hindent --validate` called vs when is regular `hindent` called?

ayazhafiz: `--validate` is called by default, as it checks that the files are properly formatted (which we want for CI).

ayazhafiz:
> Could you give me a tutorial on how to use this system you built?

Sure. In most cases you can just run a script to check whether the project is in a state validated by that script. `scripts/hindent` checks that it's formatted, `scripts/hlint` checks there's no linting errors, `scripts/squashed` is specific for pull requests and checks that the difference between your branch and master is one commit (we want to squash commits before rebasing them on master when merging a PR).

The really useful part is if a script fails, you can now pass flags to the script to (semi-)automatically fix the failures.

- `scripts/hindent --fix` will format all files
- `scripts/hlint --fix` will refactor files with linting errors, if possible. Not all errors might be automatically refactorable; a common pattern would be to run `scripts/hlint` to check for errors, `scripts/hlint --fix` to fix what can be done automatically, and then run `scripts/hlint` again and manually fix any remaining errors.
- `scripts/squashed --rebase` will interactively rebase your branch on master and autosquash fixup commits. This means that if you use `git commit --fixup=<BASE_COMMIT_SHA>` to commit subsequent changes on your branch, where `<BASE_COMMIT_SHA>` is the first commit made on your branch different from master, `scripts/squashed --rebase` will automatically squash all fixup commits into one `<BASE_COMMIT_SHA>`, satisfying the script. If you don't use fixup commits, the interactive rebase will still open a UI listing the commits on the branch, from where you can specify what commits you'd like to squash.

PR #70, opened 2020-04-30
Refine grammar types

ayazhafiz:
This commit refines types in hgb's Gingerbread grammar so to better
represent the kinds of expressions and statements available to various
grammar items.


src/Grammar.hs
```diff
@@ -72,16 +77,57 @@ instance Read Type where
       readType ('i':'n':'t':maybeBits) = TInt (readMaybe maybeBits :: Maybe Bits) True
       readType ('u':'i':'n':'t':maybeBits) = TInt (readMaybe maybeBits :: Maybe Bits) False
 
+-- Represents a named reference, either to a variable or a function.
+newtype Reference =
+  Reference
+    { name :: String
+    }
```
lukasberglund, line 84: Just curious, why do you use `newtype` instead of `data` here?

ayazhafiz: According to hlint this is the preferred convention for constructors of one type.

PR #72, opened 2020-05-01
Parse final expression in block expressions

### LINKED ISSUE #71: Fix block expressions to parse for final expression

ayazhafiz:
This commit fixes blocks to properly parse the last expression present
in them, if any. If there is no final expression, the block returns an
empty expression.

This solution is still somewhat hacky and should be fixed up.

Closes #71


src/Grammar.hs
```diff
@@ -144,10 +144,12 @@ data Expression
...
   | ConditionalExpr
       { if' :: Expression
-      , then' :: BlockExpr
-      , else' :: Maybe BlockExpr
+      , then' :: Block
+      , else' :: Maybe Block
       }
   | BooleanExpr Bool
+  | BlockExpr Block
+  | EmptyExpr
```
lukasberglund, line 152: Does this mean you can now create fully empty expressions?

src/Parser.hs
```diff
@@ -111,9 +118,9 @@ parseMultiple unitParser terminator = parseMultiple' unitParser
         (otherItems, beyond) <- parseMultiple' unitParser afterDelim
         return (parsedItem : otherItems, beyond)
 
-parseStmt :: UnitParser Statement
+parseStmt :: InjectableUnitParser Statement
```
lukasberglund, line 121: Why the name injectable?

src/Parser.hs
```diff
@@ -125,32 +132,55 @@ parseStmt tokens@(first:(second:_)) =
...
-parseExpr precedence delims tokens =
-  case symbol (head tokens) of
-    Symbol.If -> parseConditional tokens
-    _ -> do
-      (left, afterLeft) <- parsePrefix delims tokens
-      parseInfix precedence delims left afterLeft
-
-parseBlock :: UnitParser BlockExpr
+parseExpr precedence delims tokens
+  | sym `elem` delims = return (EmptyExpr, tokens)
```
lukasberglund, line 145: This implies you can now create empty expressions like just having a line with `!`. Isn't that a bad idea?

src/Parser.hs
```diff
@@ -125,32 +132,55 @@ parseStmt tokens@(first:(second:_)) =
...
+     <-
+    parseEnumeration parseStmt Symbol.StmtDelim Symbol.LBrace stmtsAndBeyond
+  let stmts' =
+        if null stmts
+          then [ExpressionStatement EmptyExpr]
+          else stmts
+  let block =
+        case last stmts' of
+          (ExpressionStatement expr) -> Block (init stmts') expr
+          _ -> Block stmts EmptyExpr
```
lukasberglund, line 182: Why not just call an error here?

src/Grammar.hs
ayazhafiz: Yes, something like `!` is valid

src/Parser.hs
ayazhafiz: Because a delimiter is "injected" in it. I would like a better name but couldn't think of one..

src/Parser.hs
ayazhafiz: Maybe? I know later on it will help us in better error handling if a statement is missing an expression. I think most languages allow it.

src/Parser.hs
ayazhafiz: What do you mean?
This should be okay, just the block returns an empty expression. For instance in the block

```
{
  a : int = 2;
}
```

the last statement is a variable decl and the block returns _

lukasberglund: Wouldn't the last statement be `_` since parseenumeration parses the line after `;` too?

src/Parser.hs
lukasberglund: Why allow statements that are missing an expression?

src/Parser.hs
ayazhafiz: no that's the last expression

ayazhafiz: statements = [decl a = 2], finalExpression = _

src/Parser.hs
ayazhafiz: So that parseEnumeration can parse something like `{ foo(); }` -> [(foo), EmpyExpr]

ayazhafiz: This could be better

ayazhafiz: But I don't know how to fix the parser right now to parse expression blocks better. If you have ideas I would really appreciate it

ayazhafiz:
@lukasberglund parsing the block with `parseEnumeration` is kind of hacky, for the reasons explained in the comment I left in the source code and because it requires an empty expression. If you have a suggestion on  a better way to do it, let me know!

src/Parser.hs
lukasberglund: Oh right.

ayazhafiz: Yes. And then the code block above in `parseBlock` moves the last statement to be the last expression, so we get stmts: [2!], finalExpression: _.

src/Parser.hs
ayazhafiz: Yes, I like that idea, but how do we handle a case like `{ foo() }`? I think the two best options are

1. Try to parse `foo()` as a statement, and fall back on parsing an expression if it fails
2. Make inferences. That is, an expression is parsed until it reaches a non-expression context (i.e. no binary operators, etc) and the caller is responsible for handling everything afterward. In this sense parsers are always responsible for only parsing there context. This also means a statement parser like `parseDecl` is responsible for validating the statement ends with `!` (or some other terminator).

What do you think?

src/Parser.hs
lukasberglund: Right, but the `_` is an `ExpressionStatement`, isn't it?

src/Parser.hs
lukasberglund: I think option two could work. It could be worth doing if it has other advantages, but IDK if it's necessary here. 

As I understand if `parseEnumeration` (or our modified version of it) is called on `{ foo ()}` it returns `[ExpressionStatement foo ()]`, upon which we can return `Block [] (foo ())`. So there isn't a problem.

src/Parser.hs
ayazhafiz: Until it is processed by line 181. Line 181 pulls the expression out from the `ExpressionStatement`.

src/Parser.hs
ayazhafiz: It has other advantages because right now `{ a : int = 1 }` is permitted, but only `{ a : int = 1; }` should be permitted.

ayazhafiz: Anyway, let's merge this PR and then implement option (2) in a separate PR. It's probably more important to start focusing on other parts of the compiler like name binding and type checking anyway.

lukasberglund: Alright. But could I try my hand at the approach I was thinking first?

Issue #69, opened 2020-04-30
Codegen


ayazhafiz:
While @lukasberglund continues work on `eval` and takes a break from hgb, I would like to begin writing codegen modules. This will place in most of the pipeline we need sans typechecking and name binding.

We can output to a number of target languages. Some I am interested in include

- [ ] assembly (x86)
- [ ] TypeScript/JavaScript (nice to have because Gingerbread is at a similar level of abstraction to this, but employs different paradigms)
- [ ] A lisp (possibly racket?)
- [ ] C

lukasberglund:
Sounds good!

PR #74, opened 2020-05-01
Add TypeScript codegen for variable declarations

ayazhafiz:
This commit lays the groundwork for TypeScript code generation by adding
support for mapping a Gingerbread grammar to a TypeScript grammar for
variable declarations, and writing the TypeScript grammar out into a
string.

As described above, there are two steps for the current codegen process:

1. Map the Gingerbread grammar into another language grammar
  - this commit introduces a `MapGrammar a b` typeclass with a method
    `\> :: a -> b` that maps a `Grammar a` to a `Grammar b`
2. Write the mapped grammar out in the target language's syntax
  - this commit introduces a `Write a` typeclass with a method
    `write :: a -> String` that writes a `Grammar a` out as string.

To validate that the generated TypeScript program is syntatically
correct, we ask the TypeScript language service to parse the generated
code and report any syntatic errors. In the future (once
binding/typechecking is in place), we could verify that the generated
code has no semantic errors as well.

Part of #69


src/Grammar.hs
```diff
@@ -51,6 +60,8 @@ data Type
   | TArray Type
   deriving (Eq)
 
+instance Grammar Type
```
lukasberglund, line 63: What does this do?

ayazhafiz: Nothing, it just defines a data type as being a grammar.

Issue #51, opened 2020-04-06
Add parser support for loops.


lukasberglund:
We need to add parser support for while loops and for loops. Here is the syntax we had in mind
```
for i in (start .. limit) {
        print("number = ", i)!
}

while start <= limit {
        print("number = ", 1 + 1)!
	start = start + 1!
}

```
I'm really annoyed that Github hasn't added Syntax highlighting support for gingerbread yet. They are really behind the curve.

ayazhafiz:
haha. we could [add it](https://github.com/github/linguist/blob/master/CONTRIBUTING.md)!

lukasberglund:
@ayazhafiz what do you think of the slight changes I made to the list syntax in the for loop?

ayazhafiz:
lgtm 👍

PR #75, opened 2020-05-01
Parse while loops

ayazhafiz:
While adding this feature, I realized we have a problem in some of the
ways we parse expressions. Most expressions (e.g. binary operators)
return the token stream after the expression when they are finished
parsing, but block expressions stop at their terminating brace and the
returned token stream includes the close brace.

```
while true { doWork() }
      ^^^^              parsed expr
           ^^^^^^^^^^^^ rest of token stream

while if true {true} else {false} { doWork() }
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^              parsed expr
                                ^^^^^^^^^^^^^^ rest of token stream

Also fails with parened expressions:
while (if true {true} else {false}) { doWork() }
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^              parsed expr
                                  ^^^^^^^^^^^^^^ rest of token stream
```

This probably isn't a big deal for now, but I think the best way going
forward for consistency would be to parse the closing token with the
expression and require that blocks end with `!` in a statement context.
This could be kinda ugly though, so I'm interested in other ideas.

```
if true {
  doSomething()
} else {
  never()
}!
```

Part of #51


src/Grammar.hs
```diff
@@ -147,7 +148,10 @@ data Expression
       , then' :: BlockExpr
       , else' :: Maybe BlockExpr
       }
-  | BooleanExpr Bool
+  | WhileExpr
+      { cond :: Expression
+      , block :: BlockExpr
+      }
```
lukasberglund, line 154: How would a while loop with an expression be evaluated? It could make an array ^^

ayazhafiz: 👍

PR #76, opened 2020-05-03
Tokenize ranges, parse for-in and range expressions

ayazhafiz:
Part of #51


src/Grammar.hs
```diff
@@ -147,8 +156,11 @@ data Expression
       , then' :: BlockExpr
       , else' :: Maybe BlockExpr
       }
-  | BooleanExpr Bool
-  deriving (Eq)
+  | ForInExpr
+      { binding :: Reference
+      , range :: Range
```
lukasberglund, line 161: Don't you think this should allow for all types of list-like objects? I thought we could have the range be an array and let `a .. b` be an array constructor.

ayazhafiz: I'll fix it up in a separate PR if that's cool. We should have an `Iterator` grammar to express this.

lukasberglund: Why isn't having an array enough? Also, I thought about this and I think the best would be for `a .. b` to act like a function that returns an array.

ayazhafiz: Because list-like things might not be arrays. consider for example an infinite iterator. What do you think? 

Do you want to take the lead on this? I think you have a better vision than I do.

PR #81, opened 2020-08-05
Add TypeScript emit for boolean expressions and arrays

ayazhafiz:
Since Lukas and I need to figure out more of what this language will
look like and how sema will actually work with Haskell, in the meantime
I am adding some more TS codegen code.

I realized one problem is how to translate a code like `if x { a } else { b }`,
which is a valid expression in GB but not in TS. I think the best
translation for this is the IIFE

```typescript
(() => { if (x) { return a; } else { return b; } })()
```

This way, we don't have to create and keep track of local variables
created in the local scope.

However, this warrants discussion of how blocks in GB actually look
like. I think we decided that blocks that end with a statement (like
`{ foo! }`) evaluate to the unit type, and blocks that end with an
expression evaluate to that expression. The translation would be finish
voidly and return the relevant expression, respectively.

Part of #69


lukasberglund:
We could also translate `if x {a} else {b}` to `x ? a : b`

ayazhafiz:
Yes, the ternary operator is a good idea too. The idea is that `if true { 1 } else { 2 }` evaluates to `1` in GB, but `if (true) { 1 } else { 2 }` is a statement in JavaScript, and doesn't evaluate to anything, and cannot be put into a place where an expression is expected. So we need to wrap the statement in an expression somehow, and it looks like the best options are a ternary or an IIFE.

lukasberglund:
Would something like this be valid in GB?
```
a = {
  foo = 1!
  foo * 3
}
```
If it is that also represents a challenge in Typescript right? So I would propose we focus on blocks that are enclosed in curly braces instead and solve the problem by doing the following:
Translate `{a}`, where a is a piece of code, into `() => {a}`, and translate unterminated expressions that are supposed to signify a return statement from `foo` to `return foo;`

ayazhafiz:
Yeah, that sounds good to me. A more generic solution is to just always rewrite the last statement accordingly and previous statements normally:

```
{ foo = 1! foo * 3 } -> (() => { let foo = 1; return foo * 3; })() // => 3
```

```
{ 1 } -> (() => { return 1; })() // => 1
```

```
{ foo = 1! foo * 3; } -> (() => { let foo = 1; foo * 3; })() // => undefined, which in GB is ()
```

PR #82, opened 2020-08-07
Transform GB blocks as TS IIFEs

ayazhafiz:
Lukas and I discussed this a bit in #81 and came to the conclusion that
the best way to proceed is to first translate block expressions to
IIFEs. For example,

```
{
  a : int = 1!
  a = a + 1!
  a
}
```

would become the TypeScript code

```ts
(() => {
  let a: number = 1;
  a = a + 1;
  return a;
})();
```

this is the most straightforward translation and will enable trivial
composition to other structures like if statements, etc.

Coming out of this, though, I am more convinced of several things that
should be changed in GB and this project:

1. Blocks should end with `!` to be treated as an expression. So in the
   example above, the GB code should look like `{ ... }!` to be a
   statement. I think this is more consistent with how all the other
   statements are specified.
2. We should add some golden text-generation tests. So create a test
   corpus, with a bunch of .gb text files, and target .ts files that we
   translate code to and verify in our tests. This will make th tests
   prettier and stuff.

Part of #69


ayazhafiz:
@lukasberglund

lukasberglund:
I like the idea of creating a text generation corpus :)

PR #92, opened 2021-01-01
Remove formatting, squashed commits checks from travis

ayazhafiz:
I have converted, I am now convinced this kind of formality is a waste
of time for smaller projects where we just want to get stuff done.
Keeping hlint bc that's probably still useful? But forcing one commit
per PR and consistent formatting isn't a big deal. I think. Or i'm just
the worst.


lukasberglund:
Or both 😈

lukasberglund (approved):

PR #93, opened 2021-01-01
Pretty-print parse output for tests

ayazhafiz:
IMO, the DSL we had for checking the parse results was kind of difficult
to read. Why not just pretty-print the parsed program directly?
That's what this commit does.

Updating this took forever ): hopefully the format doesn't need to be
changed again.


lukasberglund (approved):
Good work, you are my hero, Ayaz.

PR #94, opened 2021-01-01
Start on program analysis

ayazhafiz:


src/Analysis.hs
```diff
@@ -0,0 +1,41 @@
...
+import           Data.List
+import           Data.Map                       ( Map )
+import qualified Data.Map                      as Map
+
+-- | A @Scope@ contains all information about an items in a particular scope.
+--   This can be used to query for variable type information, function arity, etc.
+data Scope
+ = Scope { outerScope :: Maybe Scope
+         , symbols :: Map Reference Type
+         }
```
lukasberglund, line 18: I should probably use Scope for the TypeChecker right?

src/Analysis.hs
```diff
@@ -0,0 +1,41 @@
...
+
+addSymbol :: Reference -> Type -> Scope -> (Scope, [Error])
+addSymbol r t s@(Scope outer syms) =
+        if Map.member r syms then (s, [DuplicateDefinition r])
+        else (Scope outer (Map.insert r t syms), )
```
lukasberglund, line 23: Are we dissallowing polymorphisms then?

src/Analysis.hs
```diff
@@ -0,0 +1,41 @@
...
+
+collectScope :: Scope -> [Statement] -> Scope
+collectScope = foldl' addToScope
+
+checkScope :: Scope -> [Statement] -> Scope
+checkScope outerScope block =
+  let newScope = Scope outerScope Map.empty
+      scope1   = collectScope newScope block
+      check :: Statement -> [Error]
+      check (VariableDeclaration )
```
lukasberglund, line 40: I'm confused. Are you defining a function called check here? How are you doing that without `=`?

lukasberglund (commented):
I'm still a bit unsure about how all this works, if you help me out with the questions I posed bellow I'll give you more of a review.

:)

ayazhafiz:
Oops, I should have left this as WIP!

